PR #1, opened 2022-11-18
Native CacheManager

ryanemerson:
Exposes a rest endpoint on port 8080 for retrieving cache rules:

```
# Get rule key
GET /rules/{rule}/{key}

# Get all keys for a rule
GET /rules/{rule}
```

Also exposes a HotRod endpoint on port 11222 for DB Syncer operations

No authentication implemented yet as @tristantarrant is working on the design.

Requires https://github.com/gingersnap-project/debezium_poc/pull/29


manager/src/main/java/io/gingersnap/HotRodServer.java
```diff
@@ -0,0 +1,52 @@
...
+
+   static class EmbeddedServerAdminOperationHandler extends AdminOperationsHandler {
+
+      public EmbeddedServerAdminOperationHandler() {
+         super(
+               new CacheCreateTask(),
+               new CacheGetOrCreateTask(),
+               new CacheNamesTask(),
+               new CacheRemoveTask()
+         );
```
jabolina, line 49: No need for the template tasks?

manager/src/main/java/io/gingersnap/RuleResource.java
```diff
@@ -0,0 +1,38 @@
...
+   @Produces("application/json")
+   public String get(String rule, String key) {
+      key = String.format("%s:%s", rule, key);
+      return cache(rule).get(key);
+   }
+
+   @GET
+   @Path("/{rule}")
+   public Collection<String> getAllKeys(String rule) {
+      return cache(rule).keySet().stream().map(s -> s.replaceFirst("^" + rule + ":", "")).collect(Collectors.toList());
```
jabolina, line 31: Do we apply the separator here too?

jabolina (approved):
LGTM overall, just a couple of questions. The runtime feels like Quarkus' magic to me though, I guess @wburns can review it better.

wburns:
LGTM, nicely done Ryan :)
I am hoping to rip out all of the HotRodServer code though and just use the decoder for the final product just an fyi :)

manager/src/main/java/io/gingersnap/HotRodServer.java
ryanemerson: We'll only need those if the db-syncer requires templates. Once we have authentication/authorization in place the user should never have access to this admin functionality.

manager/src/main/java/io/gingersnap/RuleResource.java
ryanemerson: No I remove the separator as we just want to show the key values. Unfortunately the separator is hardcoded atm. Long term we should get the value from the Rule definition and the CacheRuleWatcher for other functionality, however in this case, once the db-syncer stores rules per cache there's no need for us to be aware of the separator when retrieving keys as we can just iterate over the cache keys directly.

ryanemerson:
> I am hoping to rip out all of the HotRodServer code though and just use the decoder for the final product just an fyi :)

+1, this is just a starting point. The binary is still 73M. There's many things that we can remove :slightly_smiling_face:

PR #5, opened 2022-12-01
Minimal HR Server

wburns:
* Basis for a reduced HR server that supports all db-sync and GET commands

Note this commit does not support SSL or the Infinispan console. We may want to integrate the console with the Quarkus one though instead?

I also have to look into supporting single port with Quarkus if that is even possible 😵


wburns:
Updated to support rest end points for get and keys. Unfortunately the default methods aren't allowing me to return byte[]. Will need to change these to prevent all the String and byte[] creations.

wburns:
Also we need to update the db-syncer to not append the rule to the key as that is easily identified as the cache name. This will reduce some copying as well. For now the rest end point is adding/stripping that.

manager/src/main/java/io/gingersnapproject/hotrod/CommandProcessor.java
```diff
@@ -0,0 +1,303 @@
...
+   }
+
+   public void ping(GingersnapHeader header) {
+      this.writeResponse(header.encoder().pingResponse(header, null, this.channel, OperationStatus.Success));
+   }
+
+   public void size(GingersnapHeader header) {
+      var cache = maps.getMaps().get(header.getCacheName());
+      var size = cache == null ? -1 : cache.estimatedSize();
+      this.writeResponse(header.encoder().unsignedLongResponse(header, null, this.channel, size));
```
wburns, line 127: Oops, this doesn't work for the -1. To be honest we will want to throw an error instead when the map isn't present. One of the features missing is we probably shouldn't define maps on a per access but have them be configured by config map solely.

wburns:
This will require https://github.com/gingersnap-project/db-syncer/pull/44 to work with the changes to the rest endpoint to only work without the prefix.

infinispan/deployment/src/main/java/io/gingersnapproject/infinispan/deployment/InfinispanExtensionProcessor.java
```diff
@@ -1,100 +1,15 @@
...
-import org.jboss.jandex.AnnotationInstance;
-import org.jboss.jandex.AnnotationTarget;
-import org.jboss.jandex.DotName;
-import org.jboss.jandex.IndexView;
-
-import io.gingersnapproject.infinispan.runtime.EmbeddedCacheManagerProducer;
-import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
```
ryanemerson: We can remove this.

ryanemerson:
Removing the `infinispan-core` and `infinispan-commons` dependencies from the extension runtime brought the native binary down to 58M for me

manager/src/main/java/io/gingersnapproject/Maps.java
```diff
@@ -0,0 +1,45 @@
...
+
+import javax.inject.Singleton;
+
+import org.infinispan.commons.marshall.WrappedByteArray;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+
+@Singleton
+public class Maps {
```
ryanemerson: I think adding a `Rules` and `Rule` class might make this code simpler to understand from the consumer perspective, something like:

```java
class Rules {
   final ConcurrentMap<String, Rule> rules = ...;
   
   class Rule {
        final Cache<WrappedByteArray, WrappedByteArray> cache = ...;
        
        byte[] get(key);
        
        Stream<byte[]> keys();
   }
}

```

manager/src/main/java/io/gingersnapproject/Maps.java
```diff
@@ -0,0 +1,45 @@
...
+   private final ConcurrentMap<String, com.github.benmanes.caffeine.cache.Cache<WrappedByteArray, WrappedByteArray>> maps = new ConcurrentHashMap<>();
+   private final ConcurrentMap<String, Cache<WrappedByteArray, List<byte[]>>> multiMaps = new ConcurrentHashMap<>();
```
ryanemerson: Do we really need to use MultiMaps for the Schema?

I'm just thinking if we utilise unique keys per entry on the db-syncer side, then we can save a lot of code/maintenance here.

wburns: We don't have to, I was debating making it part of the same map. I figured it would be simpler to do it this way though so it is more clearly defined what is a Map and what is a MultiMap. But I can be convinced to change it, I personally don't feel it will save that much code/maintenance myself and if fact will be a bit uglier as we have to use an instanceof/cast for the value.

manager/src/main/java/io/gingersnapproject/Maps.java
wburns: Sure, I only tossed this together quickly to get something working.

wburns:
> Removing the `infinispan-core` and `infinispan-commons` dependencies from the extension runtime brought the native binary down to 58M for me

Nice, I can do that in here as well then too.

Although I believe commons is being pulled in by the server though as well. And if not, I am pretty sure we are using some commons classes in ISPN. I can take a look if I can just copy them over if not.

infinispan/deployment/src/main/java/io/gingersnapproject/infinispan/deployment/InfinispanExtensionProcessor.java
wburns: I will take a double look after removing the other deps

manager/src/main/java/io/gingersnapproject/Maps.java
wburns: And the problem is the way the schema works in debezium is they are effectively appending the schema updates. We need a nice way to only write the single schema update but fetch all of them, it seemed like a natural fit for multi map.

manager/src/main/java/io/gingersnapproject/Maps.java
wburns: Actually I am not liking how Rules/Rule is working out with the code you suggested. 
I will however rename it to Rules.

I think we can rework the naming of this PR imo.

wburns: And actually the more I think about the usage with the database, I think we will want to have an memory service such as this and a database service. The latter will have to be non blocking, so we don't want to tie their APIs together to prevent allocations. And then the CommandProcessor will handle the iteraction between the two I am thinking.

wburns: I renamed it to Caches for this PR for now. I am not a fan of the abstractions this is doing currently, but I want to just get this in first and the iterate on it to be honest.

manager/src/main/java/io/gingersnapproject/hotrod/CommandProcessor.java
```diff
@@ -0,0 +1,300 @@
...
+   }
+
+   public void putAll(GingersnapHeader header, Map<byte[],byte[]> entryMap) {
+      var cache = getOrCreateMap(header.getCacheName());
+
+      try {
+         for (Map.Entry<byte[], byte[]> entry : entryMap.entrySet()) {
+            cache.put(new String(entry.getKey()), new WrappedByteArray(entry.getValue()));
+         }
```
jabolina, line 136: A todo for changing this to be an atomic operation?

jabolina (approved):
LGTM, just one comment.

manager/src/main/java/io/gingersnapproject/hotrod/CommandProcessor.java
wburns: I had debated about using putAll, but even that is not an atomic operation. The main concern is the putAll would require creating yet another map and the only benefit it would give is the presize operation if the map is very large, which we will probably never hit.

jabolina:
Merged, thanks @wburns

manager/src/main/java/io/gingersnapproject/Maps.java
ryanemerson: No objections from me if it's the simplest solution. There's no point removing this HR code to simply the cache-manager if it's going to make the db-syncer integration more complex.

PR #6, opened 2022-12-02
RHAM-84 Add OpenSearch search backend

fax4ever:
https://issues.redhat.com/browse/RHAM-84

This is the pr to add a search backend to the cache manager

The service will be used by the project after that a driver for it is present (e.g.: https://github.com/gingersnap-project/cache-manager/pull/5 for instance)

### DIFF
--- a/manager/src/main/java/io/gingersnapproject/search/SearchHit.java
+++ b/manager/src/main/java/io/gingersnapproject/search/SearchHit.java
@@ -1,0 +1,5 @@
+package io.gingersnapproject.search;
+
+public record SearchHit(String indexName, String documentId, Double score) {
+
+}
--- a/manager/src/main/java/io/gingersnapproject/search/SearchResult.java
+++ b/manager/src/main/java/io/gingersnapproject/search/SearchResult.java
@@ -1,0 +1,18 @@
+package io.gingersnapproject.search;
+
+import java.time.Duration;
+import java.util.List;
+
+public interface SearchResult {
+
+   long hitCount();
+
+   boolean hitCountExact();
+
+   List<SearchHit> hits();
+
+   boolean hitsExact();
+
+   Duration duration();
+
+}
--- a/manager/src/main/java/io/gingersnapproject/search/opensearch/CommandResponseListener.java
+++ b/manager/src/main/java/io/gingersnapproject/search/opensearch/CommandResponseListener.java
@@ -1,27 +1,25 @@
 package io.gingersnapproject.search.opensearch;
 
 import java.io.IOException;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
 
 import org.apache.http.util.EntityUtils;
 import org.elasticsearch.client.Response;
 import org.elasticsearch.client.ResponseException;
 import org.elasticsearch.client.ResponseListener;
 
-class CommandResponseListener implements ResponseListener {
+import io.smallrye.mutiny.subscription.UniEmitter;
 
-   final CompletableFuture<String> future = new CompletableFuture<>();
+record CommandResponseListener(UniEmitter<? super String> emitter) implements ResponseListener {
 
    @Override
    public void onSuccess(Response response) {
       try {
          String entity = EntityUtils.toString(response.getEntity());
-         future.complete(entity);
+         emitter.complete(entity);
       } catch (IOException exception) {
          // in this case the command has been executed server side,
          // the only problem here is to get the response message from the server
-         future.complete(exception.getMessage());
+         emitter.complete(exception.getMessage());
       }
    }
 
@@ -35,12 +33,8 @@ class CommandResponseListener implements ResponseListener {
          // TODO Verify, case by case, after the driver is integrated, the ResponseException kinds we want tolerate,
          //   completing the future normally, and the ones we don't want, reporting the exception to the caller,
          //   completing the future exceptionally.
-         future.complete(exception.getMessage());
+         emitter.complete(exception.getMessage());
       }
-      future.completeExceptionally(exception);
-   }
-
-   public CompletionStage<String> completionStage() {
-      return future;
+      emitter.fail(exception);
    }
 }
--- a/manager/src/main/java/io/gingersnapproject/search/opensearch/CommandResponseListener.java
+++ b/manager/src/main/java/io/gingersnapproject/search/opensearch/CommandResponseListener.java
@@ -1,0 +1,46 @@
+package io.gingersnapproject.search.opensearch;
+
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+import org.apache.http.util.EntityUtils;
+import org.elasticsearch.client.Response;
+import org.elasticsearch.client.ResponseException;
+import org.elasticsearch.client.ResponseListener;
+
+class CommandResponseListener implements ResponseListener {
+
+   final CompletableFuture<String> future = new CompletableFuture<>();
+
+   @Override
+   public void onSuccess(Response response) {
+      try {
+         String entity = EntityUtils.toString(response.getEntity());
+         future.complete(entity);
+      } catch (IOException exception) {
+         // in this case the command has been executed server side,
+         // the only problem here is to get the response message from the server
+         future.complete(exception.getMessage());
+      }
+   }
+
+   @Override
+   public void onFailure(Exception exception) {
+      if (exception instanceof ResponseException) {
+         /*
+          * The client tries to guess what's an error and what's not, but it's too naive.
+          * A 404 on DELETE should be accepted for instance to support idempotency.
+          */
+         // TODO Verify, case by case, after the driver is integrated, the ResponseException kinds we want tolerate,
+         //   completing the future normally, and the ones we don't want, reporting the exception to the caller,
+         //   completing the future exceptionally.
+         future.complete(exception.getMessage());
+      }
+      future.completeExceptionally(exception);
+   }
+
+   public CompletionStage<String> completionStage() {
+      return future;
+   }
+}
--- a/manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
+++ b/manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
@@ -1,0 +1,70 @@
+package io.gingersnapproject.search.opensearch;
+
+import java.util.concurrent.CompletionStage;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+import org.elasticsearch.client.Request;
+import org.elasticsearch.client.RestClient;
+import org.infinispan.commons.dataconversion.internal.Json;
+
+import io.gingersnapproject.search.SearchBackend;
+import io.gingersnapproject.search.SearchResult;
+import io.quarkus.arc.lookup.LookupIfProperty;
+
+@LookupIfProperty(name = "service.opensearch.enabled", stringValue = "true")
+@ApplicationScoped
+public class OpenSearchBackend implements SearchBackend {
+
+   @Inject
+   private RestClient restClient;
+
+   @Override
+   public CompletionStage<String> mapping(String indexName) {
+      Request request = new Request("PUT", "/" + indexName);
+
+      Json mappings = Json.object("mappings",
+            Json.object("_source",
+                  Json.object("enabled", false)));
+      request.setJsonEntity(mappings.toString());
+
+      return commandSubmit(request);
+   }
+
+   @Override
+   public CompletionStage<String> put(String indexName, String documentId, Json value) {
+      Request request = new Request("PUT", "/" + indexName + "/_doc/" + documentId);
+      request.setJsonEntity(value.toString());
+
+      return commandSubmit(request);
+   }
+
+   @Override
+   public CompletionStage<String> remove(String indexName, String documentId) {
+      Request request = new Request("DELETE", "/" + indexName + "/_doc/" + documentId);
+
+      return commandSubmit(request);
+   }
+
+   @Override
+   public CompletionStage<SearchResult> query(String sql) {
+      Request request = new Request("POST", "/_plugins/_sql");
+      request.addParameter("format", "json");
+      request.setJsonEntity(Json.object("query", sql).toString());
+
+      return submitQuery(request);
+   }
+
+   private CompletionStage<String> commandSubmit(Request request) {
+      CommandResponseListener responseListener = new CommandResponseListener();
+      restClient.performRequestAsync(request, responseListener);
+      return responseListener.completionStage();
+   }
+
+   private CompletionStage<SearchResult> submitQuery(Request request) {
+      QueryResponseListener responseListener = new QueryResponseListener();
+      restClient.performRequestAsync(request, responseListener);
+      return responseListener.completionStage();
+   }
+}
--- a/manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
+++ b/manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
@@ -1,7 +1,5 @@
 package io.gingersnapproject.search.opensearch;
 
-import java.util.concurrent.CompletionStage;
-
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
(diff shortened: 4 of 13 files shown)


manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
```diff
@@ -0,0 +1,73 @@
...
+import org.jboss.logging.Logger;
+
+import io.gingersnapproject.search.SearchBackend;
+import io.quarkus.arc.lookup.LookupIfProperty;
+
+@LookupIfProperty(name = "service.opensearch.enabled", stringValue = "true")
+@ApplicationScoped
+public class OpenSearchBackend implements SearchBackend {
+
+   private static final Logger LOG = Logger.getLogger(OpenSearchBackend.class);
```
ryanemerson: Never used.

manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
```diff
@@ -0,0 +1,73 @@
...
+   @Override
+   public CompletionStage<List<Json>> query(String sql) {
+      Request request = new Request("POST", "/_sql");
+      request.addParameter("format", "json");
+      request.setJsonEntity(JsonQueryBuilder.create(sql).build());
+
+      return submitQuery(request);
+   }
+
+   private final CompletionStage<String> commandSubmit(Request request) {
```
ryanemerson: `final` keyword is redundant

manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
```diff
@@ -0,0 +1,73 @@
...
+      CommandResponseListener responseListener = new CommandResponseListener();
+      restClient.performRequestAsync(request, responseListener);
+      return responseListener.completionStage().thenApply(CommandResponse::toString);
+   }
+
+   private final CompletionStage<List<Json>> submitQuery(Request request) {
```
ryanemerson: `final` keyword is redundant

manager/src/main/java/io/gingersnapproject/search/opensearch/JsonQueryBuilder.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.infinispan.commons.dataconversion.internal.Json;
+
+class JsonQueryBuilder {
+
+   static JsonQueryBuilder create(String sql) {
+      return new JsonQueryBuilder(sql);
+   }
+
+   private Json query;
```
ryanemerson: `final Json query`

manager/src/main/java/io/gingersnapproject/search/opensearch/JsonQueryBuilder.java
```diff
@@ -0,0 +1,21 @@
+package io.gingersnapproject.search.opensearch;
+
+import org.infinispan.commons.dataconversion.internal.Json;
+
+class JsonQueryBuilder {
```
ryanemerson: Do you envisage more fields being added to this?

I'm just thinking this seems overkill opposed to doing inline: `request.setJsonEntity(Json.object().set("query", sql).toString());`

manager/src/main/java/io/gingersnapproject/search/opensearch/CommandResponseListener.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.concurrent.CompletionStage;
+
+import org.apache.http.util.EntityUtils;
+import org.elasticsearch.client.Response;
+import org.elasticsearch.client.ResponseException;
+import org.elasticsearch.client.ResponseListener;
+
+class CommandResponseListener implements ResponseListener {
+
+   CompletableFuture<CommandResponse> future = new CompletableFuture<>();
```
ryanemerson: `final`?

manager/src/main/java/io/gingersnapproject/search/opensearch/CommandResponseListener.java
```diff
@@ -0,0 +1,47 @@
...
+
+   @Override
+   public void onSuccess(Response response) {
+      CommandResponse result = new CommandResponse(true);
```
ryanemerson: Do we need to use `CommandResponse`? or can do:

```java
class CommandResponseListener implements ResponseListener {

   final CompletableFuture<String> future = new CompletableFuture<>();

   @Override
   public void onSuccess(Response response) {
      try {
         String entity = EntityUtils.toString(response.getEntity());
         future.complete(entity);
      } catch (IOException exception) {
         future.completeExceptionally(exception);
      }
   }

   @Override
   public void onFailure(Exception exception) {
      if (exception instanceof ResponseException) {
         /*
          * The client tries to guess what's an error and what's not, but it's too naive.
          * A 404 on DELETE should be accepted for instance to support idempotency.
          */
         future.complete(exception.getMessage());
      }
      future.completeExceptionally(exception);
   }

   public CompletionStage<String> completionStage() {
      return future;
   }
}

```

manager/src/main/java/io/gingersnapproject/search/opensearch/CommandResponseListener.java
```diff
@@ -0,0 +1,47 @@
...
+   @Override
+   public void onFailure(Exception exception) {
+      if (exception instanceof ResponseException) {
+         /*
+          * The client tries to guess what's an error and what's not, but it's too naive.
+          * A 404 on DELETE should be accepted for instance to support idempotency.
+          */
+         CommandResponse result = new CommandResponse(true);
+         result.exception(exception);
+         future.complete(result);
```
ryanemerson: Does this mean that the caller of `SearchBackend` has to handle  these strings?

manager/src/main/java/io/gingersnapproject/search/opensearch/JsonQueryBuilder.java
fax4ever: > Do you envisage more fields being added to this?

yes, e.g. `fetch`. But let's start simple and maybe introduce the builder when we need.

fax4ever:
Thank you @ryanemerson for the review. I applied the changes here.
I had introduced the `CommandResponse` to handle and collect all the information about what had happened on the call and make the choice of what to report to the user at the very end of the process, but you're right we don't require it at the very moment. In case we could reintroduce it if we need it.

manager/src/main/java/io/gingersnapproject/search/opensearch/JsonQueryBuilder.java
ryanemerson: Sounds good. Let's remove this class for now then if we're not using it, there's no point keeping dead code around.

manager/src/main/java/io/gingersnapproject/search/opensearch/CommandResponseListener.java
```diff
@@ -0,0 +1,43 @@
...
+   }
+
+   @Override
+   public void onFailure(Exception exception) {
+      if (exception instanceof ResponseException) {
+         /*
+          * The client tries to guess what's an error and what's not, but it's too naive.
+          * A 404 on DELETE should be accepted for instance to support idempotency.
+          */
+         future.complete(exception.getMessage());
```
ryanemerson: As a consumer of `SearchBackend` what am I supposed to do with this? For example:

```java
      SearchBackend sb;
      sb.put("indexName", "key", "value").
            whenComplete((res, t) -> {
               // What todo with res when it is Expection#getMessage ?
            });
```

fax4ever: if you compare the `QueryRL` and the `CommandRL` you can see that we tolerate stuff for the latter we don't for the former.
In the context of a command for instance some `ResponseException` are not errors. For the moment I would say to make the future complete (not exceptionally) that is the same we do in Hibernate Search for instance.
I had introduced the `CommandResponse` object to wrap the tolerated exception, maybe we could restore it.
The problem with Elasticsearch is the fact that the driver raises some `ResponseException`s in some cases even if there is no error...

fax4ever: Another idea would be to use `CommandResponse` as command future response. The problem with that is I didn't want to propagate this Elasticsearch specificity to other clients. Maybe it is better to address it (the problem of not error `ResponseException`) once for all after we've fully integrated the driver. What do you think?

ryanemerson: > In the context of a command for instance some `ResponseException` are not errors.

Any reason that we can't handle the known cases of this in the `OpenSearchBackend` for the specific command implementations? Retrying, or taking some other action, as appropriate.

That way the consumer of `SearchBackend` doesn't need to worry about these implementation details.

fax4ever: There could be some cases in which we don’t have any errors and we shouldn’t do anything.
For instance a 404 in case of delete document does not represent necessarily an error.
My idea would be to address all these cases (one by one) after the integration with the driver has been done.
I added a todo.
The main idea of tolerating some exception that does not represent an error that is specific to the ES client is not to propagate this behavior to other search backends.

fax4ever: For sure the other kinds of exception (different from the ES ResponseException), such as `org.apache.http.ConnectionClosedException: Connection is closed` shouldn't be tolerated and the future must be complete exceptionally for them. But the ES client sometimes reports as an error what is not usually considered an error.

fax4ever:
@ryanemerson I finally added the test and evolved a bit the `SearchBackend` interface that is supposed to be used within the same Cache Manager, so not exposing it as Cache Manager API.
@wburns As soon as we have the events from the db sync on cache manager I can start to implement the api that will use this service.

manager/src/main/java/io/gingersnapproject/search/SearchHit.java
```diff
@@ -0,0 +1,13 @@
+package io.gingersnapproject.search;
+
+public record SearchHit(String indexName, String documentId, Double score) {
+
+   @Override
+   public String toString() {
```
wburns: I believe record already generates something similar, no?

manager/src/main/java/io/gingersnapproject/search/SearchResult.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.util.List;
+
+public interface SearchResult {
+
+   long hitCount();
+
+   boolean hitCountExact();
+
+   List<SearchHit> hits();
```
wburns, line 12: I worry about paging for this. I would think we would want to stream these results directly to a response object. Does ES not support that? Either way I would assume this returns a Publisher<SearchHit> or Multi<SearchHit>.

manager/src/main/java/io/gingersnapproject/search/SearchResult.java
```diff
@@ -0,0 +1,17 @@
...
+
+   boolean hitsExact();
```
wburns, line 14: How does this differ to `hitCountExact`?

manager/src/main/java/io/gingersnapproject/search/SearchResult.java
```diff
@@ -0,0 +1,17 @@
...
+
+   int duration();
```
wburns: Is this supposed to be execution time? Normally these are a long? Maybe we should also provide TimeUnit as an argument or will this always be in milliseconds or something?

manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
```diff
@@ -0,0 +1,70 @@
...
+import io.gingersnapproject.search.SearchBackend;
+import io.gingersnapproject.search.SearchResult;
+import io.quarkus.arc.lookup.LookupIfProperty;
+
+@LookupIfProperty(name = "service.opensearch.enabled", stringValue = "true")
+@ApplicationScoped
+public class OpenSearchBackend implements SearchBackend {
+
+   @Inject
+   private RestClient restClient;
```
wburns, line 20: No changes needed here, more informational question.

Do you have more information on the threading scheme for how this processes responses? I would love to resume the processing on the same netty thread that started the request for this. But I would need to see more info on it first.

wburns (commented):
I think we should start using Uni/Multi instead of CompletionStage/Publisher to be more inline with what the Quarkus guys are doing.

manager/src/main/java/io/gingersnapproject/search/SearchResult.java
fax4ever: `hitsExact` is about the accuracy of the hits, this is always true if no timeout occurs. In case of timeout Elasticsearch (or Hibernate Seerch) returns a partial list of hits and `hitsExact` will be `false`.
While `hitCountExact` is about the accuracy of the `hitCount`.
You can find more about that on [query result object topic](https://infinispan.zulipchat.com/#narrow/stream/313207-memory-service/topic/query.20result.20object).

manager/src/main/java/io/gingersnapproject/search/SearchHit.java
fax4ever: I've never used records, so I trust you.
Going to remove the toString here

manager/src/main/java/io/gingersnapproject/search/SearchResult.java
fax4ever: Yes ES (and OpenSearch) does not support that.
We could expose a streamed api using the search after (the next generation of scrolling api), but IMO it would be a stretch.
It would be the same as exposing a streaming API from a non-reactive JDBC driver. Nothing good in my opinion.
I think that it would be easier to expose a streaming API when we switch to Hibernate Search.
For the moment I would keep it as it is.

wburns: I am not sure I like this approach. When I see this method I assume it is not going to block, but from my understanding it could right? We need to make sure to be non blocking from the start and only be that way. If there is a place that is blocking we need to call that out immediately.

And long term this API needs to be non blocking, so I personally feel it is better to have it that way to begin.

manager/src/main/java/io/gingersnapproject/search/SearchResult.java
fax4ever: Good idea. Going to return a `java.time.Duration` object!

manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
fax4ever: Very interesting!
The Elasticsearch (low level -- that is the recommended) rest client uses Apache http client library (org.apache.httpcomponents:httpclient:4.5.13) and I think that the the asynchronous responses are handled by the `I/O dispatcher`s threads.
The thread (not necessarily a netty thread I think, e.g. for our test is the main Junit thread I think) that starts the request shares with the dispatcher thread the future object, so I would say it depends on how the client will handle the future I think.
I would suggest you continue the topic as soon as we integrated this service with the project APIs.
WDYT @wburns?

fax4ever:
Thank you @wburns for the review.
I applied some suggestions. Let me know if there is something else we can improve here.

fax4ever:
> I think we should start using Uni/Multi instead of CompletionStage/Publisher to be more inline with what the Quarkus guys are doing.

I tried to do that here: https://github.com/gingersnap-project/cache-manager/pull/6/commits/72aa4e0e0d8d31419fbb480f332b4684a00174a8

ryanemerson:
Thanks @fax4ever!

We can resume the Multi vs Uni debate as we progress and the needs of the endpoints become more clear, but we need to  work on exposing the query capabilities via the REST endpoint now.

fax4ever:
Makes sense to me.
Thank you @ryanemerson

manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
wburns: When I debugged they definitely came back on an I/O dispatcher thread. I think I need to just make a github issue to rework how we handle reactive responses so it is always executed on the same event loop that the request came from. This won't prevent additional context switches but at least moves all the processing of the response to the single thread.

fax4ever: @wburns is this the issue https://github.com/gingersnap-project/cache-manager/issues/17?

wburns: Yes, correct. I had forgotten to back link it to here.

PR #7, opened 2022-12-02
Use rule name to retrieve cache

jabolina:
Opening this with https://github.com/gingersnap-project/db-syncer/pull/41


jabolina:
Changed in #5

PR #14, opened 2022-12-09
Add in DB lazy load support

wburns:
Adds in required DB access that is configurable via application.properties.

Lazily loads the value from the DB if not present. It will cache null values. This change does not include expiration, which will be required for lazy cache only mode (eager does not require this).

Includes basis for weighed eviction where can estimate size based on JVM on heap. Still need to inspect JOL (Java Object Layout) to get these numbers more precise.

### DIFF
--- a/manager/pom.xml
+++ b/manager/pom.xml
@@ -41,6 +41,14 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-kubernetes-service-binding</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-reactive-mysql-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
             <artifactId>quarkus-resteasy-reactive</artifactId>
         </dependency>
         <dependency>
--- a/manager/src/main/java/io/gingersnapproject/Caches.java
+++ b/manager/src/main/java/io/gingersnapproject/Caches.java
@@ -5,40 +5,92 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.stream.Stream;
 
+import javax.inject.Inject;
 import javax.inject.Singleton;
-
-import org.infinispan.commons.marshall.WrappedByteArray;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
+
+import io.gingersnapproject.database.DatabaseHandler;
+import io.gingersnapproject.mutiny.UniItem;
+import io.smallrye.mutiny.Uni;
 
 @Singleton
 public class Caches {
-   private final ConcurrentMap<String, Cache<String, WrappedByteArray>> maps = new ConcurrentHashMap<>();
+   private final ConcurrentMap<String, LoadingCache<String, Uni<String>>> maps = new ConcurrentHashMap<>();
    private final ConcurrentMap<String, Cache<String, List<byte[]>>> multiMaps = new ConcurrentHashMap<>();
 
-   private Cache<String, WrappedByteArray> getOrCreateMap(String name) {
-      // TODO: eventually need to apply eviction to this. Can use a weigher to count byte[] size per key/value
-      return maps.computeIfAbsent(name, ___ -> Caffeine.newBuilder().build());
+   @Inject
+   DatabaseHandler databaseHandler;
+
+   public ConcurrentMap<String, Cache<String, List<byte[]>>> getMultiMaps() {
+      return multiMaps;
    }
 
-   public byte[] get(String name, String key) {
-      WrappedByteArray value = getOrCreateMap(name).getIfPresent(key);
-      return value == null ? null : value.getBytes();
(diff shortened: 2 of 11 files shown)


manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
```diff
@@ -0,0 +1,60 @@
...
+         throw new IllegalArgumentException("No rule found for " + rule);
+      }
+
+      // Have to do this until bug is fixed allowing injection of reactive Pool
+      PreparedQuery<RowSet<Row>> preparedQuery = pool.preparedQuery(ruleConfig.selectStatement());
+      var quarkusPreparedQuery = io.vertx.mutiny.sqlclient.PreparedQuery.<RowSet<Row>>newInstance(preparedQuery);
+
+//      return pool.preparedQuery(ruleConfig.selectStatement())
+      return quarkusPreparedQuery
+            .execute(Tuple.from(key.split(ruleConfig.plainSeparator())))
```
jabolina: JSON keys could raise a problem here. Or do only accept plain?

wburns: Haven't implemented JSON yet :D

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -5,40 +5,91 @@
...
+            })
+            .build(key -> {
+               Uni<String> dbUni = databaseHandler.select(name, key)
+                     // Make sure to use memoize, so that if multiple people subscribe to this it won't cause
+                     // multiple DB lookups
+                     .memoize().indefinitely();
+               // This will replace the pending Uni from the DB with a UniItem so we can properly size the entry
+               // Note due to how lazy subscribe works the entry won't be present in the map  yet
+               dbUni.subscribe()
+                     .with(result -> replace(name, key, dbUni, result));
```
wburns: If an error happens we probably should remove the dbUni from the cache as well.

manager/src/main/java/io/gingersnapproject/util/UniItem.java
```diff
@@ -0,0 +1,54 @@
+package io.gingersnapproject.util;
```
ryanemerson: Nitpick, but is `.mutiny` a more meaningful package name?

manager/pom.xml
```diff
@@ -30,6 +30,14 @@
             <artifactId>infinispan-extension</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
```
ryanemerson, line 42: We should also add the service-binding extension:

```xml
        <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-kubernetes-service-binding</artifactId>
        </dependency>
```

manager/pom.xml
```diff
@@ -30,6 +30,14 @@
...
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-reactive-mysql-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-reactive-oracle-client</artifactId>
+        </dependency>
```
ryanemerson, line 49: I'm guessing this was included for your local testing at this stage?

As `quarkus.datasource.db-kind=` is required at build time, I guess there's no point ever having multiple DB dependencies simultaneously. Are you happy to remove this for now and we can introduce additional vendor dependencies when we have per DB builds?

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -5,40 +5,98 @@
...
+      // TODO: technically this says it removed something even if the Uni contained a null value prior
+      // We put a null uni into the map if a value existed before. This way we cache the tombstone.
+      return getOrCreateMap(name).asMap()
+            .replace(key, Uni.createFrom().nullItem()) != null;
    }
 
-   public ConcurrentMap<String, Cache<String, List<byte[]>>> getMultiMaps() {
-      return multiMaps;
+   // Returns an estimated size which includes tombstone values
+   public long estimatedSize(String name) {
```
ryanemerson: There's no harm having this, but did you have a specific use-case in mind where we need to have this available via HotRod?

I assume we'll need it for /metrics.

manager/src/main/java/io/gingersnapproject/util/UniItem.java
wburns: I was planning on putting various utility things here, but sure I can change it.

manager/pom.xml
wburns: I don't think this should be in this PR though right?

manager/pom.xml
wburns: I am more than fine removing until we have a PR DB build

manager/src/main/java/io/gingersnapproject/Caches.java
wburns: I can remove, sure. It was just a simple command that I could implement.

manager/pom.xml
ryanemerson: As we're using the reactive client I don't think any additional implementation is required, just the dependency, so I think it makes sense to include it now.

wburns: Alright, as long as it is just a simple dep update :)

wburns:
Updated

ryanemerson:
Thanks @wburns

PR #18, opened 2022-12-14
Index is updated on put/remove

wburns:
Integrate with noop and opensearch providers

### DIFF
--- a/manager/src/main/java/io/gingersnapproject/Caches.java
+++ b/manager/src/main/java/io/gingersnapproject/Caches.java
@@ -17,6 +17,7 @@ import io.gingersnapproject.database.DatabaseHandler;
 import io.gingersnapproject.metrics.CacheAccessRecord;
 import io.gingersnapproject.metrics.CacheManagerMetrics;
 import io.gingersnapproject.mutiny.UniItem;
+import io.gingersnapproject.search.SearchBackend;
 import io.smallrye.mutiny.Uni;
 
 @Singleton
@@ -30,6 +31,9 @@ public class Caches {
    DatabaseHandler databaseHandler;
    @Inject
    Configuration configuration;
+
+   @Inject
+   SearchBackend searchBackend;
 
    public ConcurrentMap<String, Cache<String, List<byte[]>>> getMultiMaps() {
       return multiMaps;
@@ -62,6 +66,7 @@ public class Caches {
                      .memoize().indefinitely();
                // This will replace the pending Uni from the DB with a UniItem so we can properly size the entry
                // Note due to how lazy subscribe works the entry won't be present in the map  yet
+               // TODO: technically only want to do this on caller subscribing
                dbUni.subscribe()
                      .with(result -> replace(name, key, dbUni, result), t -> actualRemove(name, key, dbUni));
                return dbUni;
@@ -82,9 +87,17 @@ public class Caches {
       }
    }
 
-   public void put(String name, String key, String value) {
+   public Uni<String> put(String name, String key, String value) {
+      Uni<String> indexUni = searchBackend.put(name, key, value)
+                  .map(___ -> value)
+            // Make sure subsequent subscriptions don't update index again
+                  .memoize().indefinitely();
       getOrCreateMap(name)
-            .put(key, UniItem.fromItem(value));
+            .put(key, indexUni);
+      // TODO: technically only want to do this on caller subscribing
+      indexUni.subscribe()
+                  .with(s -> replace(name, key, indexUni, value), t -> actualRemove(name, key, indexUni));
+      return indexUni;
    }
 
    public Uni<String> get(String name, String key) {
@@ -106,10 +119,21 @@ public class Caches {
             .stream();
    }
 
-   public boolean remove(String name, String key) {
-      // TODO: technically this says it removed something even if the Uni contained a null value prior
+   public Uni<Boolean> remove(String name, String key) {
+      Uni<String> indexUni = searchBackend.remove(name, key)
+            .<String>map(___ -> null)
+            // Make sure subsequent subscriptions don't update index again
+            .memoize().indefinitely();
       // We put a null uni into the map if a value existed before. This way we cache the tombstone.
-      return getOrCreateMap(name).asMap()
-            .replace(key, Uni.createFrom().nullItem()) != null;
+      Uni<String> prev = getOrCreateMap(name).asMap()
+            .replace(key, indexUni);
+      if (prev != null) {
+         // TODO: technically only want to do this on caller subscribing
+         indexUni.subscribe()
+               .with(s -> replace(name, key, indexUni, null), t -> actualRemove(name, key, indexUni));
+         // TODO: technically this says it removed something even if the Uni contained a null value prior
+         return Uni.createFrom().item(Boolean.TRUE);
+      }
+      return Uni.createFrom().item(Boolean.FALSE);
    }
 }
--- a/manager/src/main/java/io/gingersnapproject/search/NoOpSearchBackend.java
+++ b/manager/src/main/java/io/gingersnapproject/search/NoOpSearchBackend.java
@@ -1,0 +1,27 @@
+package io.gingersnapproject.search;
+
+import io.quarkus.arc.lookup.LookupIfProperty;
+import io.smallrye.mutiny.Uni;
+
+@LookupIfProperty(name = SearchBackend.PROPERTY, stringValue = "none", lookupIfMissing = true)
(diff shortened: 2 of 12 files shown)


manager/src/main/java/io/gingersnapproject/search/NoOpSearchBackend.java
```diff
@@ -0,0 +1,29 @@
+package io.gingersnapproject.search;
+
+import org.infinispan.commons.dataconversion.internal.Json;
+
+import io.quarkus.arc.lookup.LookupIfProperty;
+import io.smallrye.mutiny.Uni;
+
+@LookupIfProperty(name = SearchBackend.PROPERTY, stringValue = "none", lookupIfMissing = true)
+public class NoOpSearchBackend implements SearchBackend {
```
ryanemerson, line 7: With this implementation it's only possible for Query to either be enabled or disabled for all Rules right?

I thought we wanted to make it so that we can enable/disable query per rule?

wburns: Correct, the query backend has not yet been plugged into using rules. That is the next step. This is more related to do we want more than one backend search? For example do we want both elasticsearch and hibernate search at the same time? I was thinking not, so this allows us to enable an explicit one.

manager/src/main/java/io/gingersnapproject/search/QueryHandler.java
```diff
@@ -0,0 +1,25 @@
...
+   @Inject
+   Caches caches;
+
+   @Inject
+   SearchBackend searchBackend;
+
+   public Multi<String> query(String query) {
+      Uni<SearchResult> result = searchBackend.query(query);
+      return result.toMulti().onItem().transformToIterable(SearchResult::hits)
+            .onItem().transformToUni(sh -> caches.get(sh.indexName(), sh.documentId()))
```
wburns, line 21: Next step improvement for this is we should add a `getIfPresent` method on `Caches` and for each entry that isn't present we need to do a batch query or IN clause from the DB. Currently this could cause up to N database selects if none of the entries are in the cache.

manager/src/main/java/io/gingersnapproject/search/NoOpSearchBackend.java
ryanemerson: Discussed on Zulip, but +1 on limiting to a single query impl at anyone time.

manager/src/main/java/io/gingersnapproject/search/opensearch/CommandResponseListener.java
```diff
@@ -34,7 +34,8 @@ public void onFailure(Exception exception) {
          //   completing the future normally, and the ones we don't want, reporting the exception to the caller,
          //   completing the future exceptionally.
          emitter.complete(exception.getMessage());
+      } else {
+         emitter.fail(exception);
       }
-      emitter.fail(exception);
    }
```
fax4ever, line 40: Thank you for this fix @wburns.
Just for curiosity, what is happen if we do:
> emitter.complete("...");

and then:
> emitter.fail("...");

?

manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
```diff
@@ -32,9 +33,9 @@ public Uni<String> mapping(String indexName) {
    }
 
    @Override
-   public Uni<String> put(String indexName, String documentId, Json value) {
+   public Uni<String> put(String indexName, String documentId, String jsonString) {
```
fax4ever, line 49: good to not parse it again

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -71,9 +76,17 @@ private void actualRemove(String name, String key, Uni<String> value) {
...
+      Uni<String> indexUni = searchBackend.put(name, key, value)
+                  .map(___ -> value)
+            // Make sure subsequent subscriptions don't update index again
+                  .memoize().indefinitely();
       getOrCreateMap(name)
-            .put(key, UniItem.fromItem(value));
+            .put(key, indexUni);
+      // TODO: technically only want to do this on caller subscribing
+      indexUni.subscribe()
+                  .with(s -> replace(name, key, indexUni, value), t -> actualRemove(name, key, indexUni));
```
fax4ever, line 99: Ok for the `actualRemove` in case of indexing error, but do we need the replace too?

fax4ever (approved):
just one commet
+1 to merge from me

manager/src/main/java/io/gingersnapproject/search/opensearch/CommandResponseListener.java
wburns: It should be fine as the first completion should be the one who wins, but figure might as well not be ambiguous :)

manager/src/main/java/io/gingersnapproject/Caches.java
wburns: We need the replace to work for eviction to be able to apply the weigher properly.

Basically the idea is a cached value in the Caches map will always be a UniItem so we can properly estimate its size in memory.

manager/src/main/java/io/gingersnapproject/search/opensearch/CommandResponseListener.java
fax4ever: Definitely

manager/src/main/java/io/gingersnapproject/Caches.java
fax4ever: Ok makes sense. Didn't think about eviction.

manager/src/main/java/io/gingersnapproject/search/QueryHandler.java
```diff
@@ -0,0 +1,25 @@
...
+
+@Singleton
+public class QueryHandler {
+   @Inject
+   Caches caches;
+
+   @Inject
+   SearchBackend searchBackend;
+
+   public Multi<String> query(String query) {
```
fax4ever, line 18: Maybe we should expose the hitCount too. See https://infinispan.zulipchat.com/#narrow/stream/313207-memory-service/topic/Cache-Manager/near/316208754

wburns: I would like to add that in a different PR, so this can be integrated. A bunch of other PRs hinge on this.

fax4ever: Sure. Feel free to rebase and merge from me.

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -82,9 +87,17 @@ private void actualRemove(String name, String key, Uni<String> value) {
       }
    }
 
-   public void put(String name, String key, String value) {
+   public Uni<String> put(String name, String key, String value) {
+      Uni<String> indexUni = searchBackend.put(name, key, value)
+                  .map(___ -> value)
+            // Make sure subsequent subscriptions don't update index again
+                  .memoize().indefinitely();
```
jabolina, line 94: We'll end up indexing db-syncer offset and schema. I guess we can leave that as a to-do tough.

wburns: This is handled by https://github.com/gingersnap-project/cache-manager/pull/21

manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
```diff
@@ -12,12 +15,14 @@
...
 public class OpenSearchBackend implements SearchBackend {
 
    @Inject
-   private RestClient restClient;
+   RestClient restClient;
+
+   Set<String> mappedRules = ConcurrentHashMap.newKeySet();
 
    @Override
    public Uni<String> mapping(String indexName) {
```
jabolina, line 28: Not for this PR, but I guess we'll need a way to remove a mapping.

jabolina (approved):

manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
wburns: Yeah, I figure that will come as part of the listener stuff. There are lot of issues there still that need to be ironed out. Removing map, index, clearing index etc

jabolina:
Thanks, @wburns

PR #19, opened 2022-12-15
Disable ElasticSearch health check

ryanemerson:
Add HealthCheckerTest

In the future once a Rule is marked as queryable, we should include an
additional @Readiness check to ensure that the Elasticsearch client can
communicate with the backends when one or more Rules with query support
exist.


jabolina (approved):

jabolina:
Thanks, @ryanemerson

PR #21, opened 2022-12-15
Query enable by rule

wburns:
Only allows for query operations to be performed on rules that have query enabled. If not enabled the query operation is just plainly ignored. Also if a user configures a rule with querying without a backend query provider it will error.

### DIFF
--- a/manager/src/main/java/io/gingersnapproject/Caches.java
+++ b/manager/src/main/java/io/gingersnapproject/Caches.java
@@ -17,7 +17,7 @@ import io.gingersnapproject.database.DatabaseHandler;
 import io.gingersnapproject.metrics.CacheAccessRecord;
 import io.gingersnapproject.metrics.CacheManagerMetrics;
 import io.gingersnapproject.mutiny.UniItem;
-import io.gingersnapproject.search.QueryHandler;
+import io.gingersnapproject.search.IndexingHandler;
 import io.smallrye.mutiny.Uni;
 
 @Singleton
@@ -33,7 +33,7 @@ public class Caches {
    Configuration configuration;
 
    @Inject
-   QueryHandler queryHandler;
+   IndexingHandler indexingHandler;
 
    public ConcurrentMap<String, Cache<String, List<byte[]>>> getMultiMaps() {
       return multiMaps;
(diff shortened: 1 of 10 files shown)


wburns:
Relies upon https://github.com/gingersnap-project/cache-manager/pull/18

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -26,6 +29,9 @@ public class Caches {
    @Inject
    DatabaseHandler databaseHandler;
 
+   @Inject
+   QueryHandler queryHandler;
```
fax4ever: I think that we cannot have a circular dependency here (Caches <--> QueryHandler), maybe it is better to keep to inject the SearchBackend here

manager/src/main/java/io/gingersnapproject/search/SearchBackend.java
```diff
@@ -1,14 +1,12 @@
 package io.gingersnapproject.search;
 
-import org.infinispan.commons.dataconversion.internal.Json;
-
 import io.smallrye.mutiny.Uni;
 
 public interface SearchBackend {
 
-   Uni<String> mapping(String indexName);
```
fax4ever, line 9: Good idea to hide the mapping, as we said this is not functionality

fax4ever (commented):
Nice job! Just a comment about the ci failure

fax4ever:
Here is an idea: https://github.com/gingersnap-project/cache-manager/commit/e385d58840e291fcbb45841e4b88f0e7d2f95cad

manager/src/main/java/io/gingersnapproject/Caches.java
wburns: Yeah, I didn't find time yesterday to fix this, yet.

wburns:
> Here is an idea: [e385d58](https://github.com/gingersnap-project/cache-manager/commit/e385d58840e291fcbb45841e4b88f0e7d2f95cad)

Nice, yes this is precisely what I was thinking, but didn't get time to get it. Thanks!

wburns:
> > Here is an idea: [e385d58](https://github.com/gingersnap-project/cache-manager/commit/e385d58840e291fcbb45841e4b88f0e7d2f95cad)
> 
> Nice, yes this is precisely what I was thinking, but didn't get time to get it. Thanks!

Added your commit to my PR.

wburns:
I need to revamp the tests a bit now, since it is not using the proper rule any longer and causes test failures.

fax4ever:
Merged, thanks @wburns!

PR #23, opened 2022-12-15
Added `getAll` operation

jabolina:
The `getAll` operation is used by the db-syncer to retrieve offset information.

### DIFF
--- a/manager/pom.xml
+++ b/manager/pom.xml
@@ -90,6 +90,12 @@
             <artifactId>mysql-connector-java</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.infinispan</groupId>
+            <artifactId>infinispan-client-hotrod</artifactId>
+            <version>${version.infinispan}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/manager/src/main/java/io/gingersnapproject/Caches.java
+++ b/manager/src/main/java/io/gingersnapproject/Caches.java
@@ -1,6 +1,20 @@
 package io.gingersnapproject;
 
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
 import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.CacheLoader;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import io.gingersnapproject.configuration.Configuration;
@@ -10,6 +24,8 @@ import io.gingersnapproject.metrics.CacheManagerMetrics;
 import io.gingersnapproject.mutiny.UniItem;
 import io.gingersnapproject.search.IndexingHandler;
 import io.smallrye.mutiny.Uni;
+import io.smallrye.mutiny.groups.UniJoin;
+import org.infinispan.util.KeyValuePair;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
@@ -19,6 +35,8 @@ import java.util.stream.Stream;
 
 @Singleton
 public class Caches {
+
+   private static final Uni<Map<String, String>> EMPTY_MAP_UNI = Uni.createFrom().item(Collections.emptyMap());
    @Inject
    CacheManagerMetrics metrics;
    private final ConcurrentMap<String, LoadingCache<String, Uni<String>>> maps = new ConcurrentHashMap<>();
@@ -61,6 +79,16 @@ public class Caches {
       return indexUni;
    }
 
+   public Uni<List<String>> putAll(String name, Map<String, String> values) {
+      // TODO use bulk operation to index
+      UniJoin.Builder<String> builder = Uni.join().builder();
+      for (Map.Entry<String, String> entry : values.entrySet()) {
+         builder = builder.add(put(name, entry.getKey(), entry.getValue()));
+      }
+
+      return builder.joinAll().andFailFast();
+   }
+
    public Uni<String> get(String name, String key) {
       CacheAccessRecord<String> cacheAccessRecord = metrics.recordCacheAccess();
       try {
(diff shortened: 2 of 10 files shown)


manager/pom.xml
```diff
@@ -47,6 +47,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-reactive-mysql-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-reactive-pg-client</artifactId>
+        </dependency>
```
jabolina: We'll need this when running db-syncer with Postgres during tests. Adding in an additional commit.

ryanemerson: We actually have a bit of a problem with multiple DB vendors atm. The DatabaseHandler injects a vertx `Pool` instance, however the implementation is tied to the `quarkus.datasource.db-kind` property which can only be configured at build time.

Eventually we will want different build profiles and/or modules per DB vendor, but this also complicates testing.

@wburns Any thoughts on how best to overcome this?

jabolina: DOH, I thought db-kind could be overridden at runtime :disappointed:. To get things running I am OK removing this now and pinning the db-syncer to MySQL.

wburns: This really needs to discussed outside of the context of this PR. We should just remove this change from here.

There are 2 options though I can think of. We have to have a different build per DB, or we can define multiple named providers and based on the url the user specifies at runtime we can pick out the appropriate one. The former will produce leaner and simpler code, the latter provides more flexibility.

manager/src/main/java/io/gingersnapproject/hotrod/CommandProcessor.java
```diff
@@ -111,6 +113,26 @@ public void putAll(GingersnapHeader header, Map<byte[], byte[]> entryMap) {
       }
    }
 
+   public void getAll(GingersnapHeader header, Set<byte[]> keys) {
+      try {
+         String cacheName = header.getCacheName();
+         UniJoin.Builder<KeyValuePair<byte[], byte[]>> builder = Uni.join().builder();
```
wburns: After I thought about it, this needs to be done in a different way because https://github.com/gingersnap-project/cache-manager/pull/18/files#r1049741507 will also need it. This can be fine for a short term solution though, to get functionality working.

manager/pom.xml
jabolina: I removed the dependency for now.

manager/src/main/java/io/gingersnapproject/hotrod/CommandProcessor.java
jabolina: I added the `putAll` to the `Caches` class.

jabolina:
Rebasing.

jabolina:
Rebased, just let me test again the putAll method

jabolina:
Including a change that wraps the uses of `SearchBackend` in an `Instance`.

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -100,12 +122,22 @@ public Uni<String> put(String name, String key, String value) {
       return indexUni;
    }
 
+   public Uni<List<String>> putAll(String name, Map<String, String> values) {
+      // TODO use bulk operation to index
```
fax4ever: Created https://github.com/gingersnap-project/cache-manager/issues/27

manager/pom.xml
ryanemerson: > This really needs to discussed outside of the context of this PR. We should just remove this change from here.

+1, I just wanted to make sure everyone was aware of the issue.

I have created https://github.com/gingersnap-project/cache-manager/issues/28 to continue the discussion and track the work.

fax4ever (commented):
Nice job @jabolina! LGMT
I wanted to just add a test case covering the new 
1. getAll API 
2. putAll SPI

manager/src/main/java/io/gingersnapproject/search/SearchBackendFactory.java
```diff
@@ -0,0 +1,30 @@
...
+import javax.inject.Singleton;
+
+import io.gingersnapproject.search.opensearch.OpenSearchBackend;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.elasticsearch.client.RestClient;
+
+public class SearchBackendFactory {
+
+    @ConfigProperty(name = SearchBackend.PROPERTY, defaultValue = "null")
```
wburns: Can this just not use the Instance field of the SearchBackend? I don't like doing our own factory logic here tbh. Also adding a new impl doesn't require any changes other than the new SearchBackend impl.

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -61,6 +71,16 @@ public Uni<String> put(String name, String key, String value) {
       return indexUni;
    }
 
+   public Uni<List<String>> putAll(String name, Map<String, String> values) {
```
wburns: This should be in a separate PR imo. We need to make sure to have getAll in first.

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -80,6 +100,28 @@ public Stream<String> getKeys(String name) {
...
+      CacheAccessRecord<String> cacheAccessRecord = metrics.recordCacheAccess();
+      try {
+         Map<String, Uni<String>> res = getOrCreateMap(name).getAll(keys);
+
+         if (res.isEmpty()) return EMPTY_MAP_UNI;
+
+         UniJoin.Builder<KeyValuePair<String, String>> builder = Uni.join().builder();
+         for (Map.Entry<String, Uni<String>> entry : res.entrySet()) {
+            String key = entry.getKey();
+            cacheAccessRecord.localHit(entry.getValue() instanceof UniItem<String>);
```
wburns: This won't work, we would need a separate cache access record per key.

Also this doesn't look like it completes the access. `uni.onItemOrFailure().invoke(cacheAccessRecord)`

manager/src/main/java/io/gingersnapproject/hotrod/CommandProcessor.java
```diff
@@ -162,10 +164,51 @@ public void putAll(GingersnapHeader header, Map<byte[], byte[]> entryMap) {
...
             return;
          }
+         Map<String, String> transformed = new HashMap<>();
          for (Map.Entry<byte[], byte[]> entry : entryMap.entrySet()) {
-            maps.put(cacheName, toString(entry.getKey()), toString(entry.getValue()));
+            transformed.put(toString(entry.getKey()), toString(entry.getValue()));
          }
-         writeSuccess(header);
+
+         maps.putAll(cacheName, transformed).subscribe()
```
wburns: Lets do this in a separate PR. This has a bit of noise now.

manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
```diff
@@ -3,27 +3,25 @@
...
 public class OpenSearchBackend implements SearchBackend {
+   public static final String OPENSEARCH_ENABLED = "opensearch";
 
-   @Inject
-   RestClient restClient;
+   private final RestClient restClient;
 
    Set<String> mappedRules = ConcurrentHashMap.newKeySet();
 
+   public OpenSearchBackend(RestClient restClient) {
```
wburns: These changes are unneeded if we use Instance in the wrapper class instead.

manager/src/main/java/io/gingersnapproject/hotrod/CommandProcessor.java
jabolina: Sure, I'll revert in here and open another.

manager/src/main/java/io/gingersnapproject/Caches.java
jabolina: I delegated this to the `get` operation.

jabolina:
I reverted the `putAll` change, and I'll do it on another PR.

manager/src/test/java/io/gingersnapproject/hotrod/HotRodOperationsTest.java
```diff
@@ -0,0 +1,63 @@
...
+      if (cm != null) {
+         cm.stop();
+      }
+
+      cm = null;
+      cache = null;
+   }
+
+   @Test
+   public void testPutAll() {
```
wburns:
```suggestion
   public void testGetAll() {
```

wburns (approved):
LGTM, just a simple change to test

jabolina:
The `TimerMetric` needs to use the rule name during registration. Since every rule has the same name, the `MetricsResourceTest` is failing. Since my test does some operations, the metric of another rule is affected.

wburns:
Integrated into main, the MetricsResourceTest is randomly failing for me as well. We need to fix that asap now though.

PR #24, opened 2022-12-16
Expose per rule cache size metric

pruivo:


manager/src/test/java/io/gingersnapproject/mysql/MySQLResources.java
```diff
@@ -1,43 +1,49 @@
...
+
+   private MySQLContainer<?> db;
+
+   @Override
+   public Map<String, String> start() {
+      db = new MySQLContainer<>(IMAGE)
+            .withUsername("gingersnap_user")
+            .withPassword("password")
+            .withExposedPorts(MySQLContainer.MYSQL_PORT)
+            .withFileSystemBind(HOST_TMP, CONTAINER_DATA_DIR, BindMode.READ_WRITE)
```
pruivo, line 28: be aware that I'm binding the `/tmp/` directory into the container. It makes the test way faster (I have the docker data on an HDD) and the data is there if you re-run the test.
I can remove it or use a system property to set it up.

ryanemerson: No objections from me, that's great to know it speeds things up.

manager/src/main/java/io/gingersnapproject/metrics/CacheManagerMetrics.java
```diff
@@ -15,4 +17,11 @@ public interface CacheManagerMetrics {
     */
    <T> CacheAccessRecord<T> recordCacheAccess();
 
+   /**
+    * Register metrics associated with the {@code rule} stored in {@code cache}.
+    *
+    * @param rule  The rule's name.
+    * @param cache The {@link LoadingCache} where the data is cached.
+    */
+   void registerRulesMetric(String rule, LoadingCache<?, ?> cache);
```
ryanemerson: We also need the ability to deregister a Rule's metric for when they're removed at runtime. This will be called by the ConfigMap informer.

manager/src/test/java/io/gingersnapproject/mysql/MySQLResources.java
fax4ever: Great idea IMO too. Maybe we should apply it to all tests.

ryanemerson: Sounds like a good plan to me. The faster the testsuite the better

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -112,4 +83,41 @@ public boolean remove(String name, String key) {
...
+                  // Note due to how lazy subscribe works the entry won't be present in the map  yet
+                  dbUni.subscribe()
+                        .with(result -> replace(rule, key, dbUni, result), t -> actualRemove(rule, key, dbUni));
+                  return dbUni;
+               });
+      } else {
+         // If no rule defined, don't apply database loading (internal caches)
+         cache = Caffeine.newBuilder().build(k -> null);
+      }
+      metrics.registerRulesMetric(rule, cache);
```
wburns: I am torn on the metrics for the non rule maps. This is currently just the offset, the schema is stored in the multi map.

manager/src/main/java/io/gingersnapproject/metrics/CacheManagerMetrics.java
pruivo: is it implemented? I couldn't find anything yet.

wburns: I think we need to look at everything once we have the informer. I think it is fine for now.

wburns (approved):

pruivo:
@wburns I'm not collecting the cache access nor exposing the cache size for the "internal" maps.

manager/src/main/java/io/gingersnapproject/metrics/CacheManagerMetrics.java
ryanemerson: A draft of the Informer PR is already up https://github.com/gingersnap-project/cache-manager/pull/25. This PR will land first, so if we add this additional method it should just be a case of @rigazilla calling it via his runtime configuration abstraction on a delete event.

ryanemerson:
@pruivo Rebase required.

pruivo:
@ryanemerson it looks like `main` is broken.

ryanemerson:
> @ryanemerson it looks like `main` is broken.

Argh my bad for merging. Thanks for fixing.

ryanemerson:
Thanks @pruivo

PR #25, opened 2022-12-16
watcher to informer

rigazilla:
informer for cache-manager (same as db-syncer)


rigazilla:
I've splitted the logic for lazy and eager both on k8s and cache-m side
we can consider to join them if implementation is the same.
On k8s a single informer can be start for both (selecting `withLabel` ?)
On cache side the xRuleManagement interfaces can be collapsed.

ryanemerson:
> I've splitted the logic for lazy and eager both on k8s and cache-m side
> we can consider to join them if implementation is the same.

Currently the cache-manager only has the notion of a "Rule", it doesn't differentiate between eager/cache, but we should probably change this as the two Rule types will start to diverge (query is only applicable to Eager already). @wburns WDYT?


> On k8s a single informer can be start for both (selecting `withLabel` ?)
> On cache side the xRuleManagement interfaces can be collapsed.

I would keep it simple and just use two Informers for now, I can't see it causing much overhead.

rigazilla:
closing this (included in #46 )

PR #26, opened 2022-12-17
Ensure private caches are isolated from normal ones

wburns:
Reserve ___ name for private caches
Do not create new caches if they aren't a rule or private cache


wburns:
This requires https://github.com/gingersnap-project/db-syncer/pull/54 to pass

ryanemerson:
Thanks @wburns

PR #29, opened 2022-12-19
Add put all bulk indexing

### LINKED ISSUE #27: Use bulk index on put all
To handle multiple inserts we could use the [bulk api](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html).

fax4ever:
Fixes https://github.com/gingersnap-project/cache-manager/issues/27

### DIFF
--- a/manager/src/main/java/io/gingersnapproject/Caches.java
+++ b/manager/src/main/java/io/gingersnapproject/Caches.java
@@ -70,6 +70,18 @@ public class Caches {
       return indexUni;
    }
 
+   public Uni<String> putAll(String name, Map<String, String> values) {
+      Uni<String> bulkIndexingOperation = indexingHandler.putAll(name, values)
+            .memoize().indefinitely();
+
+      for (Map.Entry<String, String> entry : values.entrySet()) {
+         getOrCreateMap(name)
+               .put(entry.getKey(), UniItem.fromItem(entry.getValue()));
+      }
+
+      return bulkIndexingOperation;
+   }
+
    public Uni<String> get(String name, String key) {
       CacheAccessRecord<String> cacheAccessRecord = metrics.recordCacheAccess(name);
       try {
--- a/manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
+++ b/manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
@@ -68,6 +68,8 @@ public class OpenSearchBackend implements SearchBackend {
             .transformToUni(___ -> {
                Request request = new Request("POST", "/_bulk");
 
+               request.addParameter("refresh", "wait_for");
+
                StringBuilder body = new StringBuilder();
                for (Map.Entry<String, String> entry : documents.entrySet()) {
                   body.append(Json.object("index", Json.object("_index", indexName, "_id", entry.getKey())));
--- a/manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
+++ b/manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
@@ -1,5 +1,6 @@
 package io.gingersnapproject.search.opensearch;
 
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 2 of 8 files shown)


fax4ever:
Rebased. MetricsResourceTest is randomly failing, see https://github.com/gingersnap-project/cache-manager/pull/23#issuecomment-1358138163.

fax4ever:
Rebased again. CI is supposed to be working now.

manager/src/main/java/io/gingersnapproject/search/opensearch/OpenSearchBackend.java
```diff
@@ -56,6 +57,32 @@ public Uni<String> put(String indexName, String documentId, String jsonString) {
...
+            .transformToUni(___ -> {
+               Request request = new Request("POST", "/_bulk");
+
+               request.addParameter("refresh", "wait_for");
+
+               StringBuilder body = new StringBuilder();
+               for (Map.Entry<String, String> entry : document.entrySet()) {
+                  body.append(Json.object("index", Json.object("_index", indexName, "_id", entry.getKey())));
+                  body.append("\n"); // using \n and not the system line separator, since the value will be used by the server VM
+                  body.append(entry.getValue());
```
jabolina: do we need to escape the line breaks in the value?

fax4ever: good question! in this case it seems that we don't, trying to escape it I get an error from the OpenSearch server.

jabolina: a quick way to simulate this in the test is when creating the elements for the `documents` Map, use the `Json#toPrettyString` instead of `Json#toString`. But TBH not sure how to tackle this. If we only escape, it would turn the value into a string instead of an object.

Maybe we leave it as it is and just return the malformed error?

fax4ever: The OpenSearch (and Elasticsearch) bulk indexing API does not expect a parsable JSON, I know that this is strage, but they are their api :) see https://opensearch.org/docs/1.2/opensearch/rest-api/document-apis/bulk/

fax4ever: I think that the payload is tested here -> https://github.com/fax4ever/gingersnap-cache-manager/blob/2c8c8449a1d47bae2516b46a905b0051c0933deb/manager/src/test/java/io/gingersnapproject/search/SearchBackendTest.java#L45

fax4ever: If I try to escape the value, the test fails.

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -70,6 +70,18 @@ public Uni<String> put(String name, String key, String value) {
       return indexUni;
    }
 
+   public Uni<String> putAll(String name, Map<String, String> values) {
```
jabolina, line 73: I think we need to return either void or the map of inserted values.

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -70,6 +70,18 @@ public Uni<String> put(String name, String key, String value) {
...
+      Uni<String> bulkIndexingOperation = indexingHandler.putAll(name, values)
+            .memoize().indefinitely();
+
+      for (Map.Entry<String, String> entry : values.entrySet()) {
+         getOrCreateMap(name)
+               .put(entry.getKey(), Uni.createFrom().item(entry.getValue()));
```
wburns:
```suggestion
               .put(entry.getKey(), UniItem.fromItem(entry.getValue()));
```
We have to make sure any completed value is a UniItem so eviction can work properly

wburns: One thing to note is that because this isn't dependent upon the bulk indexing operation, the value will be in the cache before the index update occurs. This is probably fine as it reduces a lot of allocations though.

manager/src/main/java/io/gingersnapproject/Caches.java
wburns: It is fine to return Uni<String>. The value is never used, we just avoid an additional chained Uni this way. It isn't the cleanest code, but should be fine.

wburns (commented):
LGTM, just some minor tweaks needed.

fax4ever:
Thanks @jabolina and @wburns for the review, just updated the branch.

jabolina (approved):
LGTM

wburns:
Integrated into main, thanks @fax4ever !

fax4ever:
Thank you!

PR #30, opened 2022-12-19
Timer metrics per rule

jabolina:
The timer metrics were global before. Recording the latency of all searches, etc, with this change we have timers per rule. We record the latency of rules instead of a global one.

### DIFF
--- a/manager/src/main/java/io/gingersnapproject/metrics/micrometer/CacheManagerMicrometerMetrics.java
+++ b/manager/src/main/java/io/gingersnapproject/metrics/micrometer/CacheManagerMicrometerMetrics.java
@@ -8,10 +8,13 @@ import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Timer;
 
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.EnumMap;
+import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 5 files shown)


manager/src/test/java/io/gingersnapproject/metrics/MetricsResourceTest.java
```diff
@@ -59,9 +59,8 @@ public void testMetricsEndpoint() {
       assertTimerMetricsValue(expectedCount);
       assertGaugeMetricsValue(expectedGauge, MySQLResources.RULE);
 
-      // cache error
+      // rule is never created, metrics remains the same
       given().when().get(GET_PATH, "non-existing-rule", "100000").then().statusCode(HttpStatus.SC_INTERNAL_SERVER_ERROR);
-      expectedCount.put(TimerMetric.CACHE_ERROR, "1.0");
       assertTimerMetricsValue(expectedCount);
       assertGaugeMetricsValue(expectedGauge, MySQLResources.RULE);
```
jabolina, line 65: TBH not sure how we can create a failure here :disappointed:

wburns (commented):
LGTM, @pruivo  can you take a look?

manager/src/main/java/io/gingersnapproject/metrics/micrometer/CacheManagerMicrometerMetrics.java
```diff
@@ -99,7 +99,33 @@ public void localHit(boolean localHit) {
...
+      }
+
+      public void register(LoadingCache<?, ?> cache) {
+         var metrics = Arrays.stream(PerRuleGaugeMetric.values())
+               .map(gauge -> gauge.registerRule(registry, rule, cache))
+               .toList();
+
+         meters.addAll(metrics);
+         for (TimerMetric metric : TimerMetric.values()) {
+            timers.put(metric, createTimer(registry, metric.metricName(rule), metric.description()));
```
pruivo, line 123: you need to collect the `Id`s so you can remove them.

jabolina: I added an iteration in the `RuleMetrics#unregister` method.

pruivo (approved):

pruivo:
merged! thanks @jabolina !

PR #33, opened 2022-12-23
Adding caffeine classes

jabolina:
Adding the Caffeine classes for reflection in native.

### DIFF
--- a/infinispan/deployment/src/main/java/io/gingersnapproject/infinispan/deployment/InfinispanExtensionProcessor.java
+++ b/infinispan/deployment/src/main/java/io/gingersnapproject/infinispan/deployment/InfinispanExtensionProcessor.java
@@ -6,7 +6,9 @@ import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
 
 class InfinispanExtensionProcessor {
     @BuildStep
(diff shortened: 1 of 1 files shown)


jabolina:
The native test needs the native image before testing. Updating the pull request action to package and then verify.

jabolina:
The test is taking wayyyyy too much time to complete. I'll try one approach using different profiles, so we can run some tests while building the image and then run only the native tests.

jabolina:
I did not find a way to run only tests with a specific profile, so I am using the JUnit filtering here. Now we run the package, which generates the runner and runs the tests. Then we run only the integration tests that need the native binary. But going from 2m to a 6m CI is a hard hit. Most of the time seems to be spent on creating the binary https://github.com/gingersnap-project/cache-manager/actions/runs/3787752505/jobs/6439838901#step:5:474.

Does anyone have any ideas? Could we do something different here?

wburns:
This is starting to get bloated. I would say we should break this into 3 separate PRs.

1. Adding caffeine classes for native compilation
2. Adding loadAll for caffeine
3. Running integration tests.

jabolina:
I'll leave this one for the classes and open the other PRs.

jabolina:
@wburns Updated this one to only add the Caffeine classes.

wburns:
Integrated into main, thanks @jabolina !

PR #36, opened 2023-01-03
Adding integration test

jabolina:
Simple integration test to run the native-built image. Using tags and profiles for distinct integration tests.

See: https://github.com/gingersnap-project/cache-manager/pull/33#issuecomment-1365916797

### DIFF
--- a/.github/workflows/pull_requests.yaml
+++ b/.github/workflows/pull_requests.yaml
@@ -30,4 +30,6 @@ jobs:
             ${{ runner.os }}-maven-
 
       - name: Build native executable
-        run: mvn package -Pnative
+        run: |
+          mvn package -Pnative -DexcludedGroups=integration-native
+          mvn test -Dquarkus.test.profile=integration-native -Dgroups=integration-native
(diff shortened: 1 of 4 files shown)


.github/workflows/pull_requests.yaml
```diff
@@ -30,4 +30,6 @@ jobs:
             ${{ runner.os }}-maven-
 
       - name: Build native executable
-        run: mvn package -Pnative
+        run: |
+          mvn package -Pnative -DexcludedGroups=integration-native
+          mvn test -Dquarkus.test.profile=integration-native -Dgroups=integration-native
```
wburns, line 35: Do we need to define both groups and quarkus test profile? What plugin does the groups property map to?

wburns (approved):
LGTM, just a simple question.

.github/workflows/pull_requests.yaml
jabolina: The `groups` is JUnit stuff. The issue was that I only wanted to run the native tests on the second pass, but we would need to annotate the non-native tests. At least I didn't find anything ready for quarkus, to just ignore by the profile or something :disappointed:.

Another approach is to use the `ExecutionCondition` as we do in db-syncer.

wburns:
Integrated into main, thanks @jabolina !

PR #37, opened 2023-01-06
Add support for mssql, db2, oracle & postgres

wburns:


manager/src/test/java/io/gingersnapproject/mysql/MySQLResources.java
```diff
@@ -30,10 +30,10 @@ public Map<String, String> start() {
       db.start();
 
       Map<String, String> properties = new HashMap<>( Map.of(
-            "quarkus.datasource.db-kind", "MYSQL",
-            "quarkus.datasource.username", db.getUsername(),
-            "quarkus.datasource.password", db.getPassword(),
-            "quarkus.datasource.reactive.url", String.format("mysql://%s:%d/debezium", db.getHost(), db.getMappedPort(MySQLContainer.MYSQL_PORT)),
+            "quarkus.datasource.mysql.db-kind", "MYSQL",
```
jabolina: Could we make the `db-kind` implicit or is it needed at build time?

wburns: Technically this line is not needed as it is a build time only property, so it is already implicitly defined. I will remove the line.

manager/src/main/resources/application.properties
```diff
@@ -6,10 +6,26 @@ quarkus.smallrye-openapi.info-contact-url=https://gingersnap-project.io
...
 
-quarkus.datasource.db-kind=MYSQL
-quarkus.datasource.username=gingersnap_user
-quarkus.datasource.password=password
-quarkus.datasource.reactive.url=mysql://localhost:3306/debezium
+quarkus.datasource.mysql.db-kind=MYSQL
+quarkus.datasource.mysql.health-exclude=true
+quarkus.datasource.mysql.username=gingersnap_user
+quarkus.datasource.mysql.password=password
+quarkus.datasource.mysql.reactive.url=mysql://localhost:3306/debezium
```
jabolina, line 13: I think we need to mark this one with `%dev.`. When running in a container with other DBs this one causes an exception because we have multiple URLs defined. If running during development we can just change the properties file.

wburns: Yeah, I noticed as well. However, I am not going to worry as I think we are abandoning this approach. We at least have the PR to look at a later point if we want to come back  to this though.

PR #39, opened 2023-01-09
Micrometer naming convention

### LINKED ISSUE #38: Micrometer naming convention
Micrometer [naming convection](https://github.com/micrometer-metrics/micrometer-docs/blob/main/src/docs/concepts/naming.adoc) suggests using `.` to separate words in the metrics name.

My initial implementation uses `_` and it should be fixed.

Also, I believe it makes more sense for the rule's name to be part of the tag.

pruivo:
Closes #38


ryanemerson:
Thanks @pruivo

PR #40, opened 2023-01-10
Provide per DB vendor cache-manager builds

ryanemerson:
@wburns @jabolina I've just added Postgres for now. If you think this is a good approach, we can add additional DBs in this or a subsequent PR.


manager/pom.xml
```diff
@@ -151,6 +191,25 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>${surefire-plugin.version}</version>
+                <configuration>
+                    <systemPropertyVariables>
+                        <quarkus.datasource.db-kind>${quarkus.datasource.db-kind}</quarkus.datasource.db-kind>
```
jabolina: This way we can distinguish which profile generated the report? I was worried about that for the db-syncer using multiple maven runs.

ryanemerson: To work around this I have uploaded the surefire-reports simply as a different GH artifact and made it so that the test reporter job processes and uploads both as artifacts. Unfortunately the test reporter won't work until PR is merged, but you can see it in use on the Operator as an [example](https://github.com/gingersnap-project/operator/runs/10191629234).


Unfortunately I don't think we can use surefire/failsafe executions in the cache-manager like https://github.com/gingersnap-project/db-syncer/pull/56 as we need to add DB specific dependencies for each of the different runs.

ryanemerson: @jabolina I was looking at https://github.com/gingersnap-project/db-syncer/pull/56/files and I think we could combine the two approaches.

We could have individual maven profiles for the per DB vendor builds (i.e. just including the required maven artifacts) as well as a "all" profile that includes the dependencies for all DB vendors. The "all" profile can make use of multiple surefire executions like your db-syncer PR so that we can have a single maven run.

The downside of this is that there will be some duplication in the pom, but I don't think that's a big deal.

WDYT?

jabolina: I think the "all" would be really convenient for quickly running tests in the CI or locally if necessary, while the per vendor we build only when publishing. We do not plan to publish the "all" image, right?

Another thing, I think we'll need to move stuff to the failsafe plugin to run the integration tests. I am not 100% sure we can run them with surefire.

ryanemerson: > We do not plan to publish the "all" image, right?

Right, no need to do that.



> Another thing, I think we'll need to move stuff to the failsafe plugin to run the integration tests. I am not 100% sure we can run them with surefire.

Sure, np with that.

wburns (approved):
Great work Ryan! Glad this can finally come together :)

manager/pom.xml
ryanemerson: The "all" approach won't work after all, as `quarkus.datasource.db-kind` still has to be set at build time before the surefire executions :slightly_frowning_face:

manager/src/test/java/io/gingersnapproject/postgres/PostgresResources.java
```diff
@@ -0,0 +1,43 @@
...
+                .withCopyFileToContainer(MountableFile.forClasspathResource("populate.sql"), "/docker-entrypoint-initdb.d/z_populate.sql");
+        db.start();
+    }
+
+    @Override
+    public void initProperties(Map<String, String> props) {
+        props.put("quarkus.datasource.username", db.getUsername());
+        props.put("quarkus.datasource.password", db.getPassword());
+        props.put("quarkus.datasource.reactive.url", String.format("postgresql://%s:%d/%s", db.getHost(), db.getMappedPort(PostgreSQLContainer.POSTGRESQL_PORT), db.getDatabaseName()));
+        DatabaseResourcesLifecyleManager.loadProperties("mysql/mysql-test.properties", props);
```
jabolina:
```suggestion
        DatabaseResourcesLifecyleManager.loadProperties("postgres/postgres-test.properties", props);
```

ryanemerson:
I've added Postgres and MSSQL, however Postgres is broken in many ways that are beyond the scope of this PR ... so I have disabled CI testing. In order for Postgres to work we need changes to the `PostgreSQLVendor` as well as those discovered by @wburns.

ryanemerson:
@jabolina @wburns CI is good!

jabolina (approved):
Really nice! I noticed we have some postgres stuff in the actions, I guess that is to have the CI in place already?

ryanemerson:
> Really nice! I noticed we have some postgres stuff in the actions, I guess that is to have the CI in place already?

Yeah exactly. CI won't build postgres for PRs, but dev images are still created in case we want to verify anything at the container level. Once we have fix(es) for Postgres support, we can just uncomment the Postgres steps on the GH action.

manager/pom.xml
```diff
@@ -158,6 +220,25 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>${surefire-plugin.version}</version>
+                <configuration>
```
pruivo: `maven-surefire-plugin` is configured in line 198. You may want to move this configuration there :+1:

ryanemerson: Thanks, I missed that when I rebased :slightly_smiling_face:

ryanemerson:
@pruivo I've updated the postgres container to use `withTmpFs`, but MSSQL was failing to start with it.

pruivo:
> @pruivo I've updated the postgres container to use `withTmpFs`, but MSSQL was failing to start with it.

tmpfs is not supported by MSSQL image. Found 2 issues, from 2017 and 2018, that aren't fixed yet :shrug:

manager/pom.xml
```diff
@@ -157,6 +222,15 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>17</source>
+                    <target>17</target>
+                    <compilerArgs>--enable-preview</compilerArgs>
```
wburns, line 231: Any reason we are enabling preview?

wburns (approved):
LGTM

wburns:

manager/pom.xml
ryanemerson: No this is a leftover, I've pushed a fix: https://github.com/gingersnap-project/cache-manager/commit/1775af995f3f6d7d0ea6023cfc55a1c8c787e605

PR #41, opened 2023-01-13
Remove MySQLResources bind to java.io.tmpdir

pruivo:


jabolina (approved):
Deleting the folder after the run doesn't work?

pruivo:
No, the files inside are created with the user id from the MySQL docker image (I had to use `sudo` to remove the directory locally)

ryanemerson:
Maybe we can figure out a way to prevent the user id issue in the future, but for now I think it's best to just remove the mount. Thanks @pruivo

PR #42, opened 2023-01-17
Foreign key entity expansion

tristantarrant:
How it works:
* at startup we query the database metadata for each table, retrieving PKs and FKs
* when the Hot Rod put operation is invoked, if the entity has FKs, we replace the FK ids with the fully resolved entities


manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -60,6 +65,26 @@ private void actualRemove(String name, String key, Uni<String> value) {
...
 
+   public  Uni<String> denormalizedPut(String name, String key, String value) {
+      Table table = databaseHandler.table(configuration.rules().get(name).connector().table());
+      if (!table.foreignKeys().isEmpty()) {
+         // We have foreign keys to resolve
+         Json json = Json.read(value);
+         UniJoin.Builder<Object> builder = Uni.join().builder();
+         for (ForeignKey fk : table.foreignKeys()) {
+            // TODO: handle composite fks
+            Json fkJson = json.atDel(fk.columns().get(0));
```
jabolina: The `fkJson` could be null if the user does a projection in db-syncer.

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -60,6 +65,26 @@ private void actualRemove(String name, String key, Uni<String> value) {
...
+            String fkId = fkJson.asString();
+            String fkRule = databaseHandler.tableToRuleName(fk.refTable());
+            builder.add(get(fkRule, fkId).map(Json::read).map(j -> json.set(fkRule, j)));
```
jabolina: This can count as a hit/miss in the metrics, is it an issue?

manager/src/main/java/io/gingersnapproject/Caches.java
tristantarrant: added a `null` check

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -60,6 +65,28 @@ private void actualRemove(String name, String key, Uni<String> value) {
...
+         // We have foreign keys to resolve
+         Json json = Json.read(value);
+         UniJoin.Builder<Object> builder = Uni.join().builder();
+         for (ForeignKey fk : table.foreignKeys()) {
+            // TODO: handle composite fks
+            Json fkJson = json.atDel(fk.columns().get(0));
+            if (fkJson != null) {
+               String fkId = fkJson.asString();
+               String fkRule = databaseHandler.tableToRuleName(fk.refTable());
+               builder.add(get(fkRule, fkId).map(Json::read).map(j -> json.set(fkRule, j)));
```
jabolina: Should we use the `refTable` here instead of `fkRule` in the `json.set`?

manager/src/main/java/io/gingersnapproject/Caches.java
wburns: Yeah, the more I am thinking about this. We really need to shift to key and value columns for eager mode. Then we can verify this stuff at startup of cache manager and throw an error at runtime if the db-syncer doesn't have the same columns.

manager/src/main/java/io/gingersnapproject/Caches.java
wburns: That is a good question about what to do for stats with the foreign keys... I am not sure the answer myself either.

Another very weird edge case is since these gets are not transactional, it is possible to find a foreign key but when we query the database the row doesn't exist. I think we may need to possibly think about doing a join to populate all the expanded foreign key columns as well.

manager/src/main/java/io/gingersnapproject/Caches.java
wburns: This line is also not thread safe. It is possible that we have multiple threads returning from the Uni and writing to the json variable concurrently.

manager/src/main/java/io/gingersnapproject/Caches.java
tristantarrant: I have replaced the call to `get(fkRule,fkId)` with `getOrCreateMap(fkRule).get(fkId)` which will avoid metric pollution

manager/src/main/java/io/gingersnapproject/Caches.java
tristantarrant: We probably want to copy the `Json` class over from Infinispan and make it thread-safe

manager/src/main/java/io/gingersnapproject/configuration/Rule.java
```diff
@@ -19,6 +19,9 @@ public interface Rule {
       @WithDefault("false")
       boolean queryEnabled();
 
+      @WithDefault("true")
+      boolean expandEntity();
```
tristantarrant, line 23: I've added this to control whether foreign entities needs to be expanded

wburns:
Integrated into main, thanks @tristantarrant !

PR #46, opened 2023-01-25
Rule wrapper

rigazilla:
Integration between k8s configmap rules and cache-manager application


manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,357 @@
...
+   @Override
+   public void onUpdate(ConfigMap oldObj, ConfigMap newObj) {
+      Map<String, String> oldMap = oldObj.getData();
+      Map<String, String> newMap = newObj.getData();
+      var olds = oldMap.entrySet();
+      var news = newMap.entrySet();
+
+      checkNoUpdatesOrThrow(oldMap, newMap, olds, news);
+
+      var added = new HashSet<Map.Entry<String, String>>(news);
```
ryanemerson: We can just do `new HashSet<>(...`

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,357 @@
...
+      // Get added keys. From new set remove
+      // - keys already present in old set
+      added.removeIf(arg0 -> {
+         return oldMap.containsKey(arg0.getKey());
+      });
+
+      var removed = new HashSet<Map.Entry<String, String>>(olds);
```
ryanemerson: We can just do `new HashSet<>(...`

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,357 @@
...
+      // Get removed keys. From old set remove
+      // - keys still present in new set
+      removed.removeIf((arg0 -> {
```
ryanemerson: Can be simplified to a one-liner `added.removeIf(arg0 -> oldMap.containsKey(arg0.getKey()))`

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,357 @@
...
+      Map<String, String> newMap = newObj.getData();
+      var olds = oldMap.entrySet();
+      var news = newMap.entrySet();
+
+      checkNoUpdatesOrThrow(oldMap, newMap, olds, news);
+
+      var added = new HashSet<Map.Entry<String, String>>(news);
+      // Get added keys. From new set remove
+      // - keys already present in old set
+      added.removeIf(arg0 -> {
```
ryanemerson: Can be simplified to a one-liner

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,357 @@
...
+   final LazyCacheRuleSpec lazyRule;
+
+   public LazyCacheRuleSpecAdapter(LazyCacheRuleSpec LazyRule) {
+      this.lazyRule = LazyRule;
+   }
+
+   @Override
+   public Connector connector() {
+      // TODO: how to implement this for lazy rules?
+      // We don't have a table field in the k8s spec
```
ryanemerson: This is only used by the `DatabaseHandler` which is not required for Lazy rules.

I think we have two options here:

1. Proceed with a single Rule configuration for both Eager & Lazy rules, in which case we need to add a field to differentiate at runtime between the two.
2. Create separate EagerRule and LazyRule configuration. The DatabaseHandler can simply process the EagerRule configs.

My preference is for 2, as I think the Eazy and Lazy rules will only continue to diverge as we add more features. @jabolina @pruivo @wburns WDYT?

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,357 @@
...
+
+   @Override
+   public String selectStatement() {
+      return lazyRule.getQuery();
+   }
+
+   @Override
+   public boolean queryEnabled() {
+      // TODO: no query enabled field in the spec.
+      // Is false the correct value?
```
ryanemerson: For now yes. Once we have a more definitive plan for Query beyond QuickJoin, we can consider adding this field to the CR.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,357 @@
...
+      LazyCacheRuleSpec.getDefaultInstance());
+      ConfigMapEventHandler eagerHandler = new ConfigMapEventHandler(rulesConfig,
+      EagerCacheRuleSpec.getDefaultInstance());
+
+      // Start informers
+      lazyInformer = lazyCm.inform(lazyHandler, RESYNC_PERIOD);
+      lazyInformer.start();
+      eagerInformer = eagerCm.inform(eagerHandler, RESYNC_PERIOD);
+      eagerInformer.start();
+   }
```
ryanemerson, line 88: The logic for the two informers is almost identical here. It would be good if we could abstract this to avoid duplication.

manager/src/main/java/io/gingersnapproject/configuration/Configuration.java
```diff
@@ -2,13 +2,36 @@
...
 
    @WithName("rule")
    Map<String, Rule> rules();
+   
+   default void addRule(String name, Rule rule) {
+      var ruleInMap = rules().putIfAbsent(name, rule);
+      if ( ruleInMap != null) {
+         throw new UnsupportedOperationException("Existing Rule("+name+") can't be updated.");
+      }
+      // TODO: add the rule to the application
```
ryanemerson: I think we should create a separate interface and implementation to handle runtime rules. That way we can utilise `javax.enterprise.event.Event`s to handle additions/removals in the various Beans, i.e. `DatabaseHandler`, `IndexingHandler` and `Caches, by simply observing these events.

The db-syncer `NotificationManager` takes a similar approach for various lifecycle events.

ryanemerson (changes requested):
PR is not compiling :slightly_frowning_face: 

Also, I think you need to update the Java level on your IDE to Java 17. I've added a few comments to point out where newer language features can be used, but there's too many to comment on all of them.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
rigazilla: I also guess 2 is better

pruivo: +1 to number 2

wburns: Definitely 2 from me.

manager/src/main/java/io/gingersnapproject/configuration/EagerConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+import io.gingersnapproject.k8s.RuleEvents;
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
+import io.smallrye.config.WithName;
+
+@ConfigMapping(prefix = "gingersnap")
+public interface EagerConfiguration {
+   static final Logger log = LoggerFactory.getLogger(LazyConfiguration.class);
+
+   HotRod hotrod();
```
ryanemerson: I don't think it makes sense to have the `HotRod` configuration as part of `EagerConfiguration`.

If we keep the root `Configuration` as before, we can then just add methods for eager and lazy methods, e.g.:

```java
interface Configuration {
   HotRod hotrod();
  
   @WithName("lazy-rule")
   @WithDefault("")
   Map<String, LazyRule> lazyRules();
   
   @WithName("eager-rule")
   @WithDefault("")
   Map<String, EagerRule> eageRules();

}
```

manager/src/main/java/io/gingersnapproject/configuration/LazyRule.java
```diff
@@ -4,11 +4,11 @@
 
 import io.smallrye.config.WithDefault;
 
-public interface Rule {
+public interface LazyRule {
```
ryanemerson: The `connector`, `queryEnabled` and `expandEntity` methods can be removed as these are not relevant to LazyCacheRules.

manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
```diff
@@ -32,7 +32,7 @@
 public class DatabaseHandler {
    private static final Logger log = LoggerFactory.getLogger(DatabaseHandler.class);
    @Inject
-   Configuration configuration;
+   LazyConfiguration configuration;
```
ryanemerson: We should be using the configuration for Eager rules here.

manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
```diff
@@ -57,7 +57,7 @@ public void refreshSchema() {
    }
 
    public Uni<String> select(String rule, String key) {
-      Rule ruleConfig = configuration.rules().get(rule);
+      LazyRule ruleConfig = configuration.rules().get(rule);
```
ryanemerson: Here we need to check for both Lazy and Eager rules as `select` is potentially used by both rule types.

This has made me realise that we have a bigger problem though ... we can have a Eager and Lazy cache rule with the same name. However, the rest endpoints are accessed via `/rules/{rule}/...` :thinking:

On the K8s side I could try to prevent both a EagerCacheRule and LazyCacheRule for the same Cache being created with the same name, however this definitely not expected behaviour in k8s given that the two are different resource types. There's also potential timing issues that could result in Rules of both types being created with the same name.

So I think we need to find a way to tolerate rules with the same name being created for both Eager and Lazy. Unfortunately I can't think of a way to achieve this without changing the rule endpoints to explicitly say the type, e.g. `/eager-rules/{rule}/...` and `/lazy-rules/{rule}/...`

\cc @jabolina @wburns

manager/src/main/java/io/gingersnapproject/search/IndexingHandler.java
```diff
@@ -17,26 +17,26 @@ public class IndexingHandler {
    Instance<SearchBackend> searchBackend;
 
    @Inject
-   Configuration configuration;
+   LazyConfiguration configuration;
 
    public Uni<String> put(String indexName, String documentId, String jsonString) {
-      Rule rule = configuration.rules().get(indexName);
+      LazyRule rule = configuration.rules().get(indexName);
```
ryanemerson: We need to take into account both Lazy and Eager rules here.

manager/src/main/java/io/gingersnapproject/configuration/LazyConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+import org.slf4j.LoggerFactory;
+
+import io.gingersnapproject.k8s.RuleEvents;
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
+import io.smallrye.config.WithName;
+
+@ConfigMapping(prefix = "gingersnap")
+public interface LazyConfiguration {
+   static final Logger log = LoggerFactory.getLogger(LazyConfiguration.class);
```
ryanemerson: `static final` is redundant on an interface.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,410 @@
...
+   void startWatching(@Observes StartupEvent ignore) {
+      log.info("startWatching(): begin");
+      if (client.isUnsatisfied()) {
+         log.info("Kubernetes client not found, not watching config map");
+         return;
+      }
+      try {
+         var cmName = configuration.lazyConfigMapName();
+         if (cmName.isEmpty()) {
+            throw new RuntimeException("Lazy configMap not specified, not watching config map");
```
ryanemerson, line 68: We're using Exceptions for flow control here, instead if the ConfigMap name is empty I think we can just have a DEBUG log and continue. That way we can specify only one ConfigMap type during dev.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,410 @@
...
+   }
+
+   public void stop(@Observes ShutdownEvent ignore) {
+      if (lazyInformer != null) {
+         log.info("Shutdown lazyInformer");
+         lazyInformer.close();
+      }
+
+      if (eagerInformer != null) {
+         log.info("Shutdown eagerInformer");
```
ryanemerson: I think a lot of the logs in this class need to be redefined as  DEBUG messages.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,410 @@
...
+      Resource<ConfigMap> cm = kc.configMaps().withName(configMapName);
+
+      var handler = new ConfigMapEventHandler(d);
+      return cm.inform(handler, RESYNC_PERIOD);
+   }
+
+   /** Knows how to hadle k8s events from informer
+    * Generates add and remove rules events for the cache-manager application
+    */
+   public class ConfigMapEventHandler implements ResourceEventHandler<ConfigMap> {
```
ryanemerson: Can be `static`.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,410 @@
...
+
+      /** Knows how to parse configMap and the specific event set to be fire
+       * Lazy and Eager dispatcher are defined below
+       */
+      interface RuleEventDispatcher {
+      void addEvent(String key, String value);
+      void removeEvent(String key, String value);
+   }
+
+   class LazyRuleEventDispatcher implements RuleEventDispatcher {
```
ryanemerson: Can be `static`.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,410 @@
...
+
+         @Override
+         public boolean expandEntity() {
+            // TODO Auto-generated method stub. Missing in the API
+            return false;
+         }
+      }
+   }
+
+   class EagerRuleEventDispatcher implements RuleEventDispatcher {
```
ryanemerson: Can be `static`.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,410 @@
...
+   @Inject
+   KubernetesConfiguration configuration;
+   @Inject
+   Event<RuleEvents.LazyRuleAdded> lazyEventAdd;
+   @Inject
+   Event<RuleEvents.LazyRuleRemoved> lazyEventRemove;
+   @Inject
+   Event<RuleEvents.EagerRuleAdded> eagerEventAdd;
+   @Inject
+   Event<RuleEvents.EagerRuleRemoved> eagerEventRemove;
```
ryanemerson: Splitting the Events into a dedicated class, similar to the db-syncer `NotificationManager`, would simplify the code a lot here as you would only need to pass one instance to the various dispatchers and you wouldn't need to have all of the `Event` variables defined. It could just be package-private for this use-case, but IMO it would improve readability.

manager/src/main/java/io/gingersnapproject/configuration/LazyRule.java
rigazilla: These methods are referred in some places:
- IndexingHandler
- DatabaseHandler
- Caches

Do they only work with EagerRules? Or we need to dig into case by case?

EDIT: ah, never mind, I missed the comments in the conversation

manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
```diff
@@ -48,43 +49,48 @@ void start(@Observes StartupEvent ignore, Config config) {
...
-      table2rule = configuration.rules().entrySet().stream()
+      table2rule = configuration.eagerRules().entrySet().stream()
             .collect(Collectors.toMap(t -> t.getValue().connector().table(), t -> t.getKey()));
    }
 
    public Uni<String> select(String rule, String key) {
-      Rule ruleConfig = configuration.rules().get(rule);
-      if (ruleConfig == null) {
+      LazyRule lazyRuleConfig = configuration.lazyRules().get(rule);
+      if (lazyRuleConfig != null) {
```
ryanemerson: So here we need to a select for both Lazy and Eager caches as we need to handle the scenario where the Eager cache doesn't contain an entry (updates, waiting on db-syncer update) and needs to try the DB directly.

I think if we introduce another interface, say `CommonRule` or just `Rule` that includes the fields common to both Easy and Lazy rule interfaces, it will make code like that found in the `select` method much simpler as we can just program via that interface.

manager/pom.xml
```diff
@@ -59,6 +62,11 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-micrometer-registry-prometheus</artifactId>
         </dependency>
+          <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java-util</artifactId>
+            <version>3.11.4</version>
```
ryanemerson: It would be good to define this as a property like the other versions.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
rigazilla: I've checked: I can split the class but we still need an instance for each specific event, also NotificationManager does the [same](https://github.com/gingersnap-project/db-syncer/blob/f5405abd3de433d05455af93945143d4e053280b/src/main/java/io/gingersnapproject/cdc/event/NotificationManager.java#L24-L33)

ryanemerson: Yeah we will still need to inject the event in the `NoficationManager` equivalent, but it means we have a lot less variables to pass to the various inner classes in `CacheRuleInformer` as we can just pass the manager.

rigazilla: oh didn't noticed that NotificationManager was an util class. ok

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,370 @@
...
+
+         @Override
+         public boolean queryEnabled() {
+            return eagerRule.getQuery().getEnabled();
+         }
+
+         @Override
+         public boolean expandEntity() {
+            // TODO Auto-generated method stub. Also missing in the API
+            return false;
```
ryanemerson: This should be true until we expose it in the CR.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,370 @@
...
+         public KeyType keyType() {
+            switch (eagerRule.getKey().getFormat()) {
+               case TEXT:
+                  return KeyType.TEXT;
+               case JSON:
+                  return KeyType.JSON;
+               case UNRECOGNIZED:
+                  throw new IllegalArgumentException("Unsupported KeyType");
+            }
+            throw new IllegalArgumentException("Unsupported KeyType");
```
ryanemerson: We can use the new "enhanced switch" here:

```java
            return switch (eagerRule.getKey().getFormat()) {
               case TEXT -> KeyType.TEXT;
               case JSON -> KeyType.JSON;
               case UNRECOGNIZED -> throw new IllegalArgumentException("Unsupported KeyType");
            };
```

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,370 @@
...
+                  public String schema() {
+                     return "";
+                  }
+
+                  @Override
+                  public String table() {
+                     return schemaTable[0];
+                  }
+               };
+               case 2 -> new ConnectorImplementation(schemaTable);
```
ryanemerson: I think here we either use two anonymous classes inline for both cases, or we make it so that `ConnectorImplementation` takes two strings as constructor args and we return a `new ConnectorImplementation` for each case with the appropriate values set. Mixing the two styles is confusing.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,370 @@
...
+         void sendEvent(String name, String ruleStr);
+      }
+
+      private static final Logger log = LoggerFactory.getLogger(ConfigMapEventHandler.class);
+
+      public ConfigMapEventHandler(RuleEventDispatcher d) {
+         dispatcher = d;
+      }
+
+      private void processConfigMapAndSend(Set<Map.Entry<String, String>> entries, SendEventFunc f) {
```
ryanemerson: We can remove the `SendEventFunc` interface if use a `BiConsumer` here:

```java
      private void processConfigMapAndSend(Set<Map.Entry<String, String>> entries, BiConsumer<String, String> consumer) {
         for (Entry<String, String> entry : entries) {
            consumer.accept(entry.getKey(), entry.getValue());
         }
      }

```

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,370 @@
...
+         if (changedNewValues.size() > 0) {
+            // Update rule is unsupported and this should never happens. Assuption here is
+            // that the configMap
+            // is corrupted, so we raise an exception and no events are sent to the rule
+            // manager. Exception contains rule with old and new values
+            var changedOldValues = new HashSet<Map.Entry<String, String>>(olds);
+            changedOldValues.removeIf(
+                  arg0 -> !newMap.containsKey(arg0.getKey()) || arg0.getValue().equals(newMap.get(arg0.getKey())));
+            throw new UnsupportedOperationException("Rules cannot be updated: new values " + changedNewValues.toString()
+                  + ", old values " + changedOldValues.toString());
```
ryanemerson: No need to explicitly call `tostring()`

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,370 @@
...
+         // - keys which aren't in the old set (added)
+         // - keys which are in the new set with same value (unchanged)
+         changedNewValues.removeIf(
+               arg0 -> !oldMap.containsKey(arg0.getKey()) || arg0.getValue().equals(oldMap.get(arg0.getKey())));
+         if (changedNewValues.size() > 0) {
+            // Update rule is unsupported and this should never happens. Assuption here is
+            // that the configMap
+            // is corrupted, so we raise an exception and no events are sent to the rule
+            // manager. Exception contains rule with old and new values
+            var changedOldValues = new HashSet<Map.Entry<String, String>>(olds);
```
ryanemerson: `new HashSet<>(olds);`

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,370 @@
...
+         removed.removeIf((arg0 -> newMap.containsKey(arg0.getKey())));
+
+         processConfigMapAndSend(removed, (name, rule) -> dispatcher.removeEvent(name, rule));
+         processConfigMapAndSend(added, (name, rule) -> dispatcher.addEvent(name, rule));
+      }
+      /** Checks that now updates are requested */
+      private void checkNoUpdatesOrThrow(Map<String, String> oldMap, Map<String, String> newMap,
+            Set<Entry<String, String>> olds,
+            Set<Entry<String, String>> news) {
+         var changedNewValues = new HashSet<Map.Entry<String, String>>(news);
```
ryanemerson: `new HashSet<>(news);`

ryanemerson:
The various `*-test.properties` files need to be updated to use eager or lazy rules.

manager/pom.xml
```diff
@@ -222,6 +231,65 @@
...
+                    <extraArgs>--experimental_allow_proto3_optional</extraArgs>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <inherited>false</inherited> <!-- only execute these in the parent -->
+                <executions>
+                    <execution>
+                        <id>git submodule update</id>
```
ryanemerson, line 321: This breaks the Dockerfile as `git` is not present. I think this should be disabled by default and only be activated if a specific profile is configured.

ryanemerson:
@rigazilla I am testing in k8s and I see from the logs that "calling onAdd for ConfigMap" is there, however when trying to access a Lazy rule via `/rule/{rule}`  I always get:

```java
ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (vert.x-eventloop-thread-0) HTTP Request to /rules/us-east/ failed, error id: 18c6e7bf-7232-4c98-a018-ca2252ae171a-1: java.lang.IllegalArgumentException: Rule us-east not configured
```

Have you tested this locally? I'm just wondering whether it's truly a bug in the code, or if my setup is to blame.

Related, I think it would be good to make a test for the `CacheRuleInformer` using [@WithKubernetesTestServer](https://quarkus.io/guides/kubernetes-client#testing).

manager/pom.xml
rigazilla: is it ok to disable it for native profile?
it ensures that the module is updated

rigazilla:
events are not propagated to observer method in interface, I've created a proper class, now seems to work.

> Related, I think it would be good to make a test for the `CacheRuleInformer` using [@WithKubernetesTestServer](https://quarkus.io/guides/kubernetes-client#testing).

Is it ok to implement test in a new pr? so we can merge others PR

manager/pom.xml
ryanemerson: That would work for the images, but maybe it's better for us to be more explicit. So how about keep enabled by default, but if a dedicated profile is enabled we disable this execution?

That way we can run the native profile locally with this execution, but in the image build we can activate the profile that disables it.

ryanemerson:
> Is it ok to implement test in a new pr? so we can merge others PR

Sure

manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
```diff
@@ -48,20 +48,15 @@ void start(@Observes StartupEvent ignore, Config config) {
    }
 
    public void refreshSchema() {
-      tables = configuration.rules().values().stream()
+      tables = configuration.eagerRules().values().stream()
             .map(rule -> vendor.describeTable(pool, rule.connector().table()).await().indefinitely())
```
ryanemerson, line 54: I think the `DatabaseHandler` also needs to handle Eager rules being added/removed.

On `@Observes  RuleEvents.EagerRuleAdded` we need to add an entry to `tables` and `table2rule`. Similarly, on `@Observes RuleEvents.EagerRuleRemoved` we need to remove the corresponding entries from `tables` and `table2rule`.

manager/src/main/java/io/gingersnapproject/configuration/ConfigurationListener.java
```diff
@@ -0,0 +1,50 @@
...
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.gingersnapproject.k8s.RuleEvents;
+import io.smallrye.config.SmallRyeConfig;
+
+public class ConfigurationListener {
+   SmallRyeConfig smConfig = ConfigProvider.getConfig().unwrap(SmallRyeConfig.class);
+   Configuration config = smConfig.getConfigMapping(Configuration.class);
```
ryanemerson: Can we just make this a bean and do `@Inject Configuration config`?

manager/src/main/java/io/gingersnapproject/configuration/ConfigurationListener.java
```diff
@@ -0,0 +1,50 @@
...
+
+   public <T> void removeRule(Map<String, T> map, String name) {
+      var removed = map.remove(name);
+      if ( removed == null) {
+         log.warn("Rule("+name+") not present. Not removed.");
+      }
+   }
+
+   public <T> void addRule(Map<String, T> map, String name, T rule) {
+      var ruleInMap = map.putIfAbsent(name, rule);
```
ryanemerson: Unfortunately we can't do this if no rules are defined on startup, as a `Collections$EmptyMap` is used, so on adding a rule we get:

```java
2023-02-02 12:19:57,170 ERROR [io.fab.kub.cli.inf.imp.cac.SharedProcessor] (-1241025797-pool-7-thread-1) v1/namespaces/test/configmaps failed invoking io.gingersnapproject.k8s.CacheRuleInformer$ConfigMapEventHandler@281a4f8a event handler: null: java.lang.UnsupportedOperationException
	at java.util.Collections$EmptyMap.putIfAbsent(Collections.java:4743)
	at io.gingersnapproject.configuration.ConfigurationListener.addRule(ConfigurationListener.java:44)
	at io.gingersnapproject.configuration.ConfigurationListener.onAddRule(ConfigurationListener.java:21)
	at io.gingersnapproject.configuration.ConfigurationListener_Observer_onAddRule_5dd0ceb17321bcda2b840520d1be2b68954b3d32.notify(Unknown Source)
	at io.quarkus.arc.impl.EventImpl$Notifier.notifyObservers(EventImpl.java:326)
	at io.quarkus.arc.impl.EventImpl$Notifier.notify(EventImpl.java:308)
	at io.quarkus.arc.impl.EventImpl.fire(EventImpl.java:76)
	at io.gingersnapproject.k8s.NotificationManager.lazyRuleAdded(NotificationManager.java:22)
	at io.gingersnapproject.k8s.NotificationManager_ClientProxy.lazyRuleAdded(Unknown Source)
	at io.gingersnapproject.k8s.CacheRuleInformer$LazyRuleEventDispatcher.addEvent(CacheRuleInformer.java:209)
	at io.gingersnapproject.k8s.CacheRuleInformer$ConfigMapEventHandler.lambda$onAdd$0(CacheRuleInformer.java:134)
	at io.gingersnapproject.k8s.CacheRuleInformer$ConfigMapEventHandler.processConfigMapAndSend(CacheRuleInformer.java:127)
	at io.gingersnapproject.k8s.CacheRuleInformer$ConfigMapEventHandler.onAdd(CacheRuleInformer.java:134)
	at io.gingersnapproject.k8s.CacheRuleInformer$ConfigMapEventHandler.onAdd(CacheRuleInformer.java:114)
	at io.fabric8.kubernetes.client.informers.impl.cache.ProcessorListener$AddNotification.handle(ProcessorListener.java:103)
	at io.fabric8.kubernetes.client.informers.impl.cache.ProcessorListener.add(ProcessorListener.java:50)
	at io.fabric8.kubernetes.client.informers.impl.cache.SharedProcessor.lambda$distribute$0(SharedProcessor.java:87)
	at io.fabric8.kubernetes.client.informers.impl.cache.SharedProcessor.lambda$distribute$1(SharedProcessor.java:110)
	at io.fabric8.kubernetes.client.utils.internal.SerialExecutor.lambda$execute$0(SerialExecutor.java:58)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.lang.Thread.run(Thread.java:833)
	at com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:705)
	at com.oracle.svm.core.posix.thread.PosixPlatformThreads.pthreadStartRoutine(PosixPlatformThreads.java:202)

```

ryanemerson: A possible solution is to re-purpose this class as a `RuleManager` bean that  keeps track of Lazy and Eazy rule configurations. On startup the manager will read all statically defined rules from the configuration, then at runtime it will use the `@Observes` events to add/remove runtime rules.

The various components that depend on rules will have to `@Inject` the `RuleManager` instead of `Configuration` and we can use `@Priority` to ensure that the `RuleManager` always receives updates first.

ryanemerson:
It would also be really helpful if we had DEBUG logs whenever a specific Rule is added/removed from the Configuration. Atm we just have when `onAdd` is called in the `CacheRuleInformer`.

ryanemerson:
I have added a static rule to avoid the `UnsupportedOperationException`, but I always get:

```java
ERROR [org.jbo.res.rea.com.cor.AbstractResteasyReactiveContext] (vert.x-eventloop-thread-0) Request failed: java.lang.IllegalArgumentException: Rule e2e-rule not configured
lazy-cache-wcnrw cache-manager 	at io.gingersnapproject.Caches.createLoadingCache(Caches.java:172)

```

manager/src/main/java/io/gingersnapproject/configuration/ConfigurationListener.java
rigazilla: yeah, working on this design

manager/src/main/java/io/gingersnapproject/configuration/ConfigurationListener.java
rigazilla: for any reason it doesn't inject anything this way

ryanemerson: Have you tried declaring the class as a Bean with `@ApplicationScoped`?

rigazilla: yeah, even with the refactoring it doesn't get injected, there must be some trick around

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,347 @@
...
+      @Override
+      public String selectStatement() {
+         var valueSelect = (eagerRule.getValue() != null && eagerRule.getValue().getValueColumnsCount() > 0)
+               ? String.join(",", eagerRule.getValue().getValueColumnsList())
+               : "*";
+         var selectStmn = "select " + valueSelect + " from " + eagerRule.getTableName();
+         if (eagerRule.getKey().getKeyColumnsCount() > 0) {
+            var whereClauseKeys = (String[]) eagerRule.getKey().getKeyColumnsList().stream().map((s) -> s + " = ?")
+                  .toArray();
+            selectStmn += "where " + String.join("and ", whereClauseKeys);
```
ryanemerson: The generated select statement fails for two reasons:

1. There's a missing space between the `eagerRule.getTableName()` and  "WHERE" clause
2. The `whereClauseKeys` assignment fails with a `ClassCastException` as it should be:

```java
            var whereClauseKeys = eagerRule.getKey().getKeyColumnsList().stream().map((s) -> s + " = ?")
                  .toArray(String[]::new);
```

When discovering ^ issues I reworked locally to test and to use a `StringBuilder`, see below:

```java
      @Override
      public String selectStatement() {
         StringBuilder sb = new StringBuilder();
         sb.append("SELECT ")
                 .append(
                         eagerRule.getValue().getValueColumnsCount() > 0
                                 ? String.join(", ", eagerRule.getValue().getValueColumnsList())
                                 : "*"
                 )
                 .append(" FROM ")
                 .append(eagerRule.getTableName());
         if (eagerRule.getKey().getKeyColumnsCount() > 0) {
            sb.append(" WHERE ").append(
                    eagerRule.getKey()
                            .getKeyColumnsList()
                            .stream()
                            .map(s -> s + " = ?")
                            .collect(Collectors.joining("AND "))
            );
         }
         return sb.toString();
      }
```

Currently every call to `selectStatement()` will result in the `String` being recreated. Instead we should only do this once by lazily initialising the String on `selectStatement` call and storing the created value as a class variable.

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,347 @@
...
+         this.eagerRule = eagerRule;
+      }
+
+      @Override
+      public Connector connector() {
+         String[] schemaTable = eagerRule.getTableName().split("\\.");
+         if (schemaTable.length > 2) {
+            throw new IllegalArgumentException("Unsupported schema format (must be table or schema.table)");
+         }
+         return new Connector() {
```
ryanemerson: We're creating a new Connector() on every method call, which is effectively every endpoint request, instead we should only do this once and return the same instance.

ryanemerson (commented):
Using the Select statement changes I commented I can get this working with both Eager and Lazy rules. However, it only works with a JVM based image. The native binary seems to have an issue with the generated Rule classes and I always get the following for Eager rules:

```java
2023-02-03 18:13:51,083 ERROR [io.fab.kub.cli.inf.imp.cac.SharedProcessor] (-685361914-pool-9-thread-1) v1/namespaces/test/configmaps failed invoking io.gingersnapproject.k8s.CacheRuleInformer$EagerEventHandler@3b2254ad event handler: null: java.lang.NullPointerException
	at io.gingersnapproject.database.DatabaseHandler.addRuleEvent(DatabaseHandler.java:69)
	at io.gingersnapproject.database.DatabaseHandler_Observer_addRuleEvent_127b01457d6b0871c879279238cc00fbc6c6376b.notify(Unknown Source)
	at io.quarkus.arc.impl.EventImpl$Notifier.notifyObservers(EventImpl.java:326)
	at io.quarkus.arc.impl.EventImpl$Notifier.notify(EventImpl.java:308)
	at io.quarkus.arc.impl.EventImpl.fire(EventImpl.java:76)
	at io.gingersnapproject.configuration.NotificationManager.eagerRuleAdded(NotificationManager.java:34)
	at io.gingersnapproject.configuration.NotificationManager_ClientProxy.eagerRuleAdded(Unknown Source)
	at io.gingersnapproject.configuration.RuleManager.addEagerRule(RuleManager.java:78)
	at io.gingersnapproject.configuration.RuleManager_ClientProxy.addEagerRule(Unknown Source)
	at io.gingersnapproject.k8s.CacheRuleInformer$EagerEventHandler.addRule(CacheRuleInformer.java:232)
	at io.gingersnapproject.k8s.CacheRuleInformer$ConfigMapEventHandler.lambda$onAdd$0(CacheRuleInformer.java:138)
	at io.gingersnapproject.k8s.CacheRuleInformer$ConfigMapEventHandler.processConfigMapAndSend(CacheRuleInformer.java:131)
	at io.gingersnapproject.k8s.CacheRuleInformer$ConfigMapEventHandler.onAdd(CacheRuleInformer.java:138)
	at io.gingersnapproject.k8s.CacheRuleInformer$ConfigMapEventHandler.onAdd(CacheRuleInformer.java:115)
	at io.fabric8.kubernetes.client.informers.impl.cache.ProcessorListener$AddNotification.handle(ProcessorListener.java:103)
	at io.fabric8.kubernetes.client.informers.impl.cache.ProcessorListener.add(ProcessorListener.java:50)
	at io.fabric8.kubernetes.client.informers.impl.cache.SharedProcessor.lambda$distribute$0(SharedProcessor.java:87)
	at io.fabric8.kubernetes.client.informers.impl.cache.SharedProcessor.lambda$distribute$1(SharedProcessor.java:110)
	at io.fabric8.kubernetes.client.utils.internal.SerialExecutor.lambda$execute$0(SerialExecutor.java:58)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.lang.Thread.run(Thread.java:833)
	at com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:705)
	at com.oracle.svm.core.posix.thread.PosixPlatformThreads.pthreadStartRoutine(PosixPlatformThreads.java:202)

```

manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -0,0 +1,386 @@
...
+                              ? String.join(", ", eagerRule.getValue().getValueColumnsList())
+                              : "*")
+                  .append(" FROM ")
+                  .append(eagerRule.getTableName());
+            if (eagerRule.getKey().getKeyColumnsCount() > 0) {
+               sb.append(" WHERE ").append(
+                     eagerRule.getKey()
+                           .getKeyColumnsList()
+                           .stream()
+                           .map(s -> s + " = ?")
```
ryanemerson, line 356: I found another bug. This will only work with a DB that uses `?` as a placeholder, e.g. MySQL. We need DB vendor specific select statements.

Let's not delay this PR any longer though, I have created https://github.com/gingersnap-project/cache-manager/issues/56 so this can be addressed in a subsequent PR.

ryanemerson:
Thanks @rigazilla

PR #47, opened 2023-01-26
Rules use dev profile

jabolina:
This is causing issues when running the image and defining different rules pointing to the same DB.


jabolina:
Just to be clear with the issue description, when pointing to the same _table_ in the same DB. Since the properties define a rule automatically.

ryanemerson:
Thanks @jabolina

PR #48, opened 2023-01-30
Test query on denormalized data

fax4ever:


ryanemerson:
Thanks @fax4ever

fax4ever:
Thank you!

manager/src/test/java/io/gingersnapproject/rules/JoinTest.java
```diff
@@ -51,8 +52,14 @@ public void afterAll() {
...
 
    @Test
-   public void testJoin() {
+   public void testJoin() throws Exception {
       cache.put("3", "{\"name\":\"BA0666\", \"scheduled_time\":\"12:00:00\", \"airline_id\": \"1\", \"gate_id\":\"3\"}");
       given().when().get(GET_PATH, "flight", "3").then().body(containsString("British Airways")).body(containsString("B1"));
+
+      Thread.sleep(1000); // we need to wait for the ES near real time behavior
+
+      given().queryParam("query", "select * from flight where name = 'BA0666' order by scheduled_time")
```
wburns, line 61: I am not sure the value of this query. It is just querying the normal flight table as far as I can see and doesn't validate the results. Also I would suggest moving this to a different test method instead of using the one that is explicitly testing a join across tables.

PR #49, opened 2023-01-31
Adding key and value column support for cache manager

wburns:
* MySQL and MSSQLServer both work

### DIFF
--- a/manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
+++ b/manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
@@ -1,10 +1,9 @@
 package io.gingersnapproject.database;
 
-import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
-import java.util.Objects;
-import java.util.function.Function;
-import java.util.stream.Collectors;
+import java.util.Optional;
 
 import javax.enterprise.event.Observes;
 import javax.inject.Inject;
@@ -17,6 +16,7 @@ import org.slf4j.LoggerFactory;
 
 import io.gingersnapproject.configuration.Configuration;
 import io.gingersnapproject.configuration.Rule;
+import io.gingersnapproject.database.model.Column;
 import io.gingersnapproject.database.model.Table;
 import io.gingersnapproject.database.vendor.Vendor;
 import io.quarkus.runtime.StartupEvent;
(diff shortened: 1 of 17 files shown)


wburns:
Have to do some more work for postgres, but figured might as well let CI and others have a look.


Also want to at add support for unique constraint so that key columns chosen are a defined unique constraint and if not to at least log a WARN message to users.

manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
```diff
@@ -48,12 +49,37 @@ void start(@Observes StartupEvent ignore, Config config) {
...
-      tables = configuration.rules().values().stream()
-            .map(rule -> vendor.describeTable(pool, rule.connector().table()).await().indefinitely())
-            .filter(Objects::nonNull)
-            .collect(Collectors.toMap(t -> t.name(), Function.identity()));
-      table2rule = configuration.rules().entrySet().stream()
-            .collect(Collectors.toMap(t -> t.getValue().connector().table(), t -> t.getKey()));
+      for (Map.Entry<String, Rule> entry : configuration.rules().entrySet()) {
+         String ruleName = entry.getKey();
+         Rule rule = entry.getValue();
+         String tableName = rule.connector().table();
```
jabolina, line 55: This got me thinking. Seems like the `table` requirement would only make sense if not using `selectStatement` because it could be anything, not bound only to the configured table.

jabolina (approved):
LGTM :+1:

manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
ryanemerson: The `table` field is only relevant to Eager rules and https://github.com/gingersnap-project/cache-manager/pull/46 should make the distinction clearer.

wburns: Correct, the problem is various things use it right now, but I would like to make it completely optional. My goal after this was to have it be mutually exclusive between

1. sql-statement (required)
2. key-columns (optional)
or
1. table (required)
2. key-columns (optional)
3. value-columns (optional)

wburns:
This needs to be updated to support Oracle now that https://github.com/gingersnap-project/cache-manager/pull/51 is integrated. However, still waiting on https://github.com/gingersnap-project/cache-manager/pull/46 since it will also have conflicts.

manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
wburns: Also the FK expansion should only be configurable when using table configuration option and not when using sql-statements.

PR #51, opened 2023-02-02
Oracle Database Support

pruivo:
To-do list:
- [x] Create GH issue
- [x] Add oracle profile to GH actions
- [ ] ~~Oracle does not have a `time` data type, only `date` or `datetime` (modify the test suite to use `datetime`?)~~
- [x] Oracle reports columns and table names in upper case (testing nightmare!)


manager/src/main/java/io/gingersnapproject/database/vendor/VendorUtil.java
```diff
@@ -0,0 +1,25 @@
...
+import io.vertx.sqlclient.Row;
+import io.vertx.sqlclient.RowSet;
+
+/**
+ * TODO!
+ */
+public enum VendorUtil {
+   ;
+
+   public static PrimaryKey extracPrimaryKey(RowSet<Row> rows) {
```
ryanemerson: I know this is a copy+paste, but there's a type :slightly_smiling_face:  `extractPrimaryKey`

manager/src/main/java/io/gingersnapproject/database/vendor/VendorUtil.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.List;
+
+import io.gingersnapproject.database.model.PrimaryKey;
+import io.vertx.sqlclient.Row;
+import io.vertx.sqlclient.RowSet;
+
+/**
+ * TODO!
+ */
+public enum VendorUtil {
```
ryanemerson: Why an `enum`?

pruivo: No particular reason. It is just an alternative way of creating a private class with a private constructor to prevent any instantiation.

pruivo: moved the method to `Vendor` interface.

pruivo:
Updated:
* Added GH actions

ryanemerson:
Thanks @pruivo

PR #55, opened 2023-02-08
Denormalized puts fail if a foreign key is not part of the value json…

ryanemerson:
It would be nice if we could handle this in the `Vendor` implementations, opposed to per put, however we don't expose the configured "values" columns in the cache-manager configuration. Therefore, in order to obtain this info we need to retrieve additional meta data from the DB via the "select-statement".

This is a quick and easy fix that should unblock @karesti work on the demo.


wburns:
Technically the select sql version should not be using FK associations. You can do whatever you want in the select, so it should be a mutually exclusive thing eventually.

wburns:
We can integrate this now, and I will rework it in https://github.com/gingersnap-project/cache-manager/pull/49 to add a check to not allow FK expansion when using a sql statement as it should only be done with table configuration.

wburns:

PR #57, opened 2023-02-08
mssql select stmt special generation

### LINKED ISSUE #56: CacheRuleInformer creates MySQL specific Select statements
The `CacheRuleInformer#EagerCacheRuleSpecAdapter` currently creates a Select statement which uses the `?` placeholder. This won't work for the other supported databases.

rigazilla:
fix #56 

adding mssql select generation

### DIFF
--- a/manager/src/main/java/io/gingersnapproject/database/vendor/Vendor.java
+++ b/manager/src/main/java/io/gingersnapproject/database/vendor/Vendor.java
@@ -55,7 +55,5 @@ public interface Vendor {
       return new PrimaryKey(pkName, pkColumns);
    }
 
-   default String whereClause(List<String> keys) {
-      return keys.stream().map(s -> s + " = ?").collect(Collectors.joining(" AND "));
-   }
+   String whereClause(List<String> keys);
 }
--- a/manager/src/main/java/io/gingersnapproject/database/vendor/Vendor.java
+++ b/manager/src/main/java/io/gingersnapproject/database/vendor/Vendor.java
@@ -2,6 +2,7 @@ package io.gingersnapproject.database.vendor;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import io.gingersnapproject.database.model.PrimaryKey;
 import io.gingersnapproject.database.model.Table;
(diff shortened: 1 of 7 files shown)


manager/src/main/java/io/gingersnapproject/k8s/CacheRuleInformer.java
```diff
@@ -370,17 +359,47 @@ public boolean queryEnabled() {
...
+               .append(" FROM ")
+               .append(eagerRule.getTableName());
+         if (eagerRule.getKey().getKeyColumnsCount() > 0) {
+            sb.append(" WHERE ").append(
+                  whereClause());
+         }
+         selectStmt = sb.toString();
+      }
+
+      private String whereClause() {
```
ryanemerson: We already have the `Vendor` interface and implementations for DB specific logic.  I think we should add a method to that interface that takes the keys/values as args and produces the select statement. We can then just call `Vendor.fromDbKind` in this adaptor.

manager/src/main/java/io/gingersnapproject/database/vendor/Vendor.java
```diff
@@ -53,4 +54,8 @@ static PrimaryKey extractPrimaryKeyInLowerCase(RowSet<Row> rows) {
       }
       return new PrimaryKey(pkName, pkColumns);
    }
+
+   default String whereClause(List<String> keys) {
```
ryanemerson: I'm not sure a `default` implementation makes sense given that the placeholders are different for every DB vendor.

Related, could you also add the implementations for Oracle and Postgres? :angel:

rigazilla: oh sure, I thought mssql was the only one with different placeholder

wburns: Sadly I already did this in https://github.com/gingersnap-project/cache-manager/pull/49/files#diff-0fd5204bef5db21cd685c49abd1953b797863596f39ca8e8f95fb349e78d2baaR16. I had done it where each vendor only overrides a method to define the parameter with position.

wburns:
Changes look okay to me. Can you double check @ryanemerson ?

ryanemerson:
Thanks @rigazilla

PR #58, opened 2023-02-08
Avoid DBhandler to run with non initialized fields

rigazilla:
This avoid that Informer starts to send events before dbhandler is started. Setting @Priority seems not enough.
I've changed all the fields initializ from startup to lazy.


ryanemerson:
Thanks @rigazilla

PR #59, opened 2023-02-09
Reuse http server for gRPC

ryanemerson:


manager/src/test/java/io/gingersnapproject/metrics/MetricsResourceTest.java
```diff
@@ -109,7 +109,7 @@ private static String metricName(String metricName, String suffix, String value,
          name += suffix;
       }
       // tags
-      name += "{%s=\"%s\",%s=\"%s\",}".formatted(
+      name += "{%s=\"%s\",%s=\"%s\"}".formatted(
```
ryanemerson, line 112: @pruivo FYI, it seems micrometer changed there formatting in a more recent version.

PR #60, opened 2023-02-09
Update git submodule before image builds

ryanemerson:


.github/workflows/build-images.yaml
```diff
@@ -39,6 +39,9 @@ jobs:
           echo "names='mysql postgres mssql oracle'" >> $GITHUB_OUTPUT
           echo "tag=${{ inputs.tag }}" >> $GITHUB_OUTPUT
 
+      - name: Checkout latest API
+        run: git submodule update --init --recursive
+
```
rigazilla: there's an option of the checkout action for this.
Should be something like:
```
      - uses: actions/checkout@v3
        with:
          ref: ${{ inputs.ref }}
          repository: ${{ inputs.repository }}
          submodule: 'recursive'
          
          see https://github.com/actions/checkout

PR #62, opened 2023-02-09
Return 404 response when Rules and/or key is unknown

ryanemerson:


manager/src/main/java/io/gingersnapproject/RuleResource.java
```diff
@@ -47,4 +59,9 @@ public Multi<String> getAllKeys(String rule) {
    public Uni<QueryResult> query(@QueryParam("query") String query) {
       return queryHandler.query(query);
    }
+
+   private void checkRuleExists(String rule) {
+      if (ruleManager.getEagerOrLazyRuleByName(rule) == null)
+         throw new NotFoundException();
```
wburns: Have to admit, I am glad this works, because this seems pretty dirty throwing an exception instead of returning a Multi completed with the exception. Leave it as is though since it is simpler and test  confirms it works still.

manager/src/main/java/io/gingersnapproject/RuleResource.java
```diff
@@ -23,19 +25,29 @@ public class RuleResource {
...
    @Produces(MediaType.APPLICATION_JSON)
    public Uni<String> get(String rule, String key) {
-      return maps.get(rule, key);
+      checkRuleExists(rule);
+      return maps.get(rule, key)
+              .onItem()
+              .invoke(v -> {
+                 if (v == null || v.isEmpty())
+                    throw new NotFoundException();
+              });
```
wburns: This should allow the happy path to have no additional allocations on our part.
```suggestion
      checkRuleExists(rule);
      Uni<String> uni = maps.get(rule, key);
      if (uni instanceof UniItem uniItem)  {
         if (unitItem.getItem() == null) {
            throw new NotFoundException();
         }
         return uni;
      }
      return uni.onItem()
         .invoke(v -> {
                 if (v == null || v.isEmpty())
                    throw new NotFoundException();
              });
```

wburns: Also I wonder should empty String be a 404? That sounds like a 204 in that case no? Either way it needs to be consistent in the code above, I have it inconsistent right now.

wburns (changes requested):

manager/src/main/java/io/gingersnapproject/RuleResource.java
ryanemerson: > That sounds like a 204 in that case no?

If we're going to support empty values, then yes.

Luckily Quarkus handles this case for us, so I have removed this check.

manager/src/main/java/io/gingersnapproject/RuleResource.java
```diff
@@ -23,19 +21,34 @@ public class RuleResource {
...
    @Operation(summary = "Retrieve a cache entry associated with the provided rule and key")
    @Path("/{rule}/{key}")
    @Produces(MediaType.APPLICATION_JSON)
    public Uni<String> get(String rule, String key) {
-      return maps.get(rule, key);
+      checkRuleExists(rule);
+      Uni<String> uni = maps.get(rule, key);
+      if (uni instanceof UniItem<?> uniItem)  {
+         if (uniItem.getItem() == null)
+            throw new NotFoundException();
```
wburns: Looks good to me, I wonder if we should cache this exception. It is probably a somewhat common case to have a miss, so I think it might be worth it.

ryanemerson: Added.

wburns (approved):
LGTM, waiting for CI.

wburns:
Looks like it is failing on oracle now :(

ryanemerson:
> Looks like it is failing on oracle now :(

More fun with column cases :roll_eyes:. I've fixed by forcing lower-case with quotes so that we don't need to modify the tests themselves.

wburns:

PR #63, opened 2023-02-09
Remove eager/lazy cache

rigazilla:
Eager cache needs to be removed from the cache map
... also lazy needs the same

### DIFF
--- a/manager/src/main/java/io/gingersnapproject/Caches.java
+++ b/manager/src/main/java/io/gingersnapproject/Caches.java
@@ -15,13 +15,14 @@ import javax.inject.Singleton;
 
 import org.infinispan.commons.dataconversion.internal.Json;
 import org.infinispan.util.KeyValuePair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -176,6 +178,17 @@ public Uni<Boolean> remove(String name, String key) {
       return Uni.createFrom().item(Boolean.FALSE);
    }
 
+   public void removeCache(String ruleName) {
```
ryanemerson: Can't we just do `@Observes RuleEvents.EagerRuleRemoved ev` here and avoid calling a method from `DatabaseHandler`?

rigazilla: Yeah maybe yes, let me check that the exec order doesn't matter.

rigazilla: seems safe, changed to use events

ryanemerson:
Thanks @rigazilla.

I'll rebase https://github.com/gingersnap-project/cache-manager/pull/62 so that I can update the returned status codes.

PR #71, opened 2023-02-17
Don't set k8s configuration in default application.properties

ryanemerson:
We need to disable this by default so that calls to `docker run ...` and `quarkus dev`  don't fail when k8s is not available.


jabolina (approved):

jabolina:
Merged, thanks @ryanemerson

PR #72, opened 2023-02-20
Add ServiceBinding Startup and Readiness health check

ryanemerson:
This is necessary to prevent cache pod(s) being marked as Ready=True before the ServiceBinding operator has updated the cache Deployment/Daemonset to include the binding mount.


ryanemerson:
Self merging as required by https://github.com/gingersnap-project/operator/pull/79

PR #73, opened 2023-02-20
Utilise ubi-micro as native runtime image

ryanemerson:
tldr; Revert in the future, but necessary evil for docker-compose health checks which require `curl`. 

This significantly increases the image size, 252MB vs 187MB, however it
provides common utilities such as `curl`. In the future we should
utilise the `quarkus-micro-image` again and copy required dependencies
to the runtime image as required.


jabolina:
> In the future we should utilise the quarkus-micro-image again and copy required dependencies to the runtime image as required.

Do you think an issue is necessary to track that?

ryanemerson:
Yes, very good point. I have created https://github.com/gingersnap-project/cache-manager/issues/74.

PR #78, opened 2023-02-22
Denormalize Eager rule db retrievals

ryanemerson:
@rigazilla I had to remove the lazy initialisation of the DatabaseHandler table data as this logic blocks the thread pool. I think changing all the dependent beans from `@Singleton` to `@ApplicationScoped` has fixed the issues you experienced with `DatabaseHandler#start()` ordering, but would

### DIFF
--- a/manager/src/main/java/io/gingersnapproject/Caches.java
+++ b/manager/src/main/java/io/gingersnapproject/Caches.java
@@ -14,6 +14,7 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Observes;
 import javax.inject.Inject;
 
+import io.gingersnapproject.configuration.Rule;
 import org.infinispan.commons.dataconversion.internal.Json;
 import org.infinispan.util.KeyValuePair;
 import org.slf4j.Logger;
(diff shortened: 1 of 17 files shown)


manager/src/test/java/io/gingersnapproject/hotrod/HotRodOperationsTest.java
```diff
@@ -48,15 +48,12 @@ public void afterEach() {
...
-      values.put("key2", "value3");
-
-      cache.put("key1", "value1");
-      cache.put("key3", "value3");
-      cache.put("key2", "value3");
-
+      var values = Map.of(
+              "key1", "{\"fullname\": \"value1\"}",
+              "key2", "{\"fullname\": \"value2\"}",
+              "key3", "{\"fullname\": \"value3\"}"
```
ryanemerson, line 54: I updated the values to JSON as `Caches#denormalize` now fails when the value is not JSON. IMO this is acceptable as we only expect a JSON to be put from the db-syncer and we always create a JSON object when reading from the DB directly.

wburns:
Looking at this the problem is we have a Singleton dependent upon an ApplicationScoped bean. Everything relies upon the DatabaseHandler being up as far as I know, so it should really be Singleton instead, no reason to proxy its initialization until later.

manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
```diff
@@ -40,56 +43,41 @@ public class DatabaseHandler {
...
-   private Vendor getVendor() {
-      if (vendor == null) {
-         log.debug("DB Kind is {}", dbKind);
-         vendor = Vendor.fromDbKind(dbKind);
-      }
-      ;
-      return vendor;
+   void start(@Observes  @Priority(10) StartupEvent ignore) {
+      vendor = Vendor.fromDbKind(dbKind);
+      refreshSchema();
```
wburns, line 51: 👍 No reason to lazily do this.

ryanemerson:
@wburns Yeah I think being consistent with either `ApplicationScoped` or `Singleton` will fix the problem. I went with `ApplicationScoped` because it will allow us to use mocks in ours tests (eventually) like we do with the db-syncer testsuite:

https://quarkus.io/guides/cdi#applicationscoped-and-singleton-look-very-similar-which-one-should-i-choose-for-my-quarkus-application

WDYT?

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -72,35 +73,37 @@ private void actualRemove(String name, String key, Uni<String> value) {
...
+         // TODO: handle composite fks
+         String fkColumn = fk.columns().get(0);
+         if (json.at(fkColumn) == null)
+            continue;
+
+         Json fkJson = json.atDel(fkColumn);
+         if (fkJson != null) {
+            denormalize = true;
+            String fkId = fkJson.asString();
+            String fkRule = databaseHandler.tableToRuleName(fk.refTable());
```
wburns, line 98: Looks like `fkRule` can be null in some cases according to https://infinispan.zulipchat.com/#narrow/stream/313207-memory-service/topic/demo/near/329448627

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -72,35 +73,37 @@ private void actualRemove(String name, String key, Uni<String> value) {
...
+            builder.add(getOrCreateMap(fkRule).get(fkId).map(Json::read).map(j -> json.set(fkColumn, j)));
```
wburns: We should probably separate this to separate lines for readability and better for stack traces if an exception happens.

manager/src/main/java/io/gingersnapproject/Caches.java
```diff
@@ -222,15 +225,27 @@ private LoadingCache<String, Uni<String>> createLoadingCache(String ruleName) {
...
+                               .onItem()
+                               .ifNotNull()
+                               .transformToUni(value -> denormalize(ruleName, value))
+                               .memoize()
+                               .indefinitely();
+                   } else {
+                       dbUni = databaseHandler.select(rule, key)
+                               .memoize()
+                               .indefinitely();
+                   }
```
wburns, line 250:
```java
dbUni = databaseHandler.select(rule, key);
if (rule instanceof EagerRule) {
   dbUni = dbUni.onItem()
         .ifNotNoll().transformToUni(value -> denormalize(ruleName, value));
}
// Make sure to use memoize, so that if multiple people subscribe to this it won't cause
// multiple DB lookups
dbUni = dbUni.memoize().indefinitely();
```

manager/src/main/java/io/gingersnapproject/Caches.java
ryanemerson: In order for denormalisation to work on the FK columns it's necessary for an Eager rule to be defined for all of the FKs being resolved. For now, while these have to be defined manually, I think the only thing we can do is produce a more meaningful message via an `IllegalStateException`.

wburns: This brings up something I believe I may have mentioned before, should we have a way to define what fks we denormalize? If not defined we would do all throw an IllegalStateException. TBH, this should be trackable at initialization time and be a configuration issue now that it is not lazily done.

ryanemerson: I think we need something. At the minute you can skip a FK column by just not including it in the value list, but I'm sure there will be scenarios where the user wants some of the FK id(s) without the normalisation.

One thing that makes this more complex is that the configuration is driven by k8s and the CRs, so it's possible that there's a delay between each of the CRs being applied in the rule-manager.

manager/src/main/java/io/gingersnapproject/Caches.java
ryanemerson: We can't do that as the `dbUni` has to be effectively final in order for it to be usable in the `dbUni.subscribe().with(...)` lambdas

wburns: You can just add an extra line to assign dbUni to a new variable. But I leave up to you if the result looks better. I do like the distinction that the new code provides in that you see exactly what an eager rule applies.

ryanemerson:
Merging, the E2E testsuite failures are not caused by these changes

PR #81, opened 2023-02-22
Plain separator is a string literal

jabolina:
The `String#split` argument is a regex string, meaning that the default separator `|` will split the key in every character. Using the `Pattern#quote` to create a string literal instead.


ryanemerson:
Thanks @jabolina

PR #90, opened 2023-02-23
Fix json returned from query REST entripoint

fax4ever:
There are stuff we can do very much better...


manager/src/test/java/io/gingersnapproject/rules/JoinTest.java
```diff
@@ -56,9 +58,12 @@ public void testJoin() throws Exception {
 
       Thread.sleep(1000); // we need to wait for the ES near real time behavior
 
+      StringExtractorMatcher stringExtractor = new StringExtractorMatcher();
       given().queryParam("query", "select * from flight where name = 'BA0666' order by scheduled_time")
             .when().get(RULE_PATH)
-            .then().body(containsString("hitCount=1"));
+            .then().body(stringExtractor);
```
ryanemerson: We should be able to just reference the json fields directly via body, e.g:

```java
.body("hitCount", equalTo(1))
.body("hits[0].name", equalTo("BA0666"));
```

ryanemerson: I've added a commit to your fork :+1:

manager/src/main/java/io/gingersnapproject/search/QueryHandler.java
```diff
@@ -24,9 +24,7 @@ public Uni<QueryResult> query(String query) {
             // Should we allow concurrency?
             .concatenate();
 
-      return result
-            .onItem()
-            .transform(searchResult -> new QueryResult(searchResult.hitCount(), searchResult.hitCountExact(),
-                  hits, searchResult.hitsExact()));
+      return Uni.combine().all().unis(result, hits.collect().asList()).combinedWith((searchResult, strings) ->
+            new QueryResult(searchResult.hitCount(), searchResult.hitCountExact(), strings, searchResult.hitsExact()));
```
wburns: This code should probably be
```java
   public Uni<QueryResult> query(String query) {
      Uni<SearchResult> result = searchBackend.get().query(query);
      return result.onItem().transformToUni(sr ->
         Multi.createFrom().iterable(sr.hits())
               .onItem().transformToUni(sh -> caches.get(sh.indexName(), sh.documentId()))
               .concatenate()
               .collect().asList()
               .onItem().transform(list -> new QueryResult(sr.hitCount(), sr.hitCountExact(), list, sr.hitsExact()))
      );
   }
```
as it describes the relationship a bit better. The combine operator is more for separate Unis, where as the hits are not available until the result Uni has completed which is more clear with this code.

manager/src/main/java/io/gingersnapproject/json/JsonHelper.java
```diff
@@ -0,0 +1,38 @@
...
+         result.set("hitsExacts", queryResult.hitsExacts());
+
+         return result.toString();
+      });
+   }
+
+   public static Json make(List<String> hits) {
+      Json array = Json.array();
+      for (String item : hits) {
+         // TODO this is very inefficient
```
wburns, line 33: I don't think there is any other way to do this if you want a nested Json object. It needs to be parsed. Or are you thinking of creating a larger String array Json representation and calling to parse once?

manager/src/main/java/io/gingersnapproject/json/JsonHelper.java
```diff
@@ -0,0 +1,38 @@
...
+import io.smallrye.mutiny.Uni;
+
+public final class JsonHelper {
+
+   private JsonHelper() {
+   }
+
+   public static Uni<String> json(Uni<QueryResult> queryResultUni) {
+      return queryResultUni.onItem().transform(queryResult -> {
+         // Json.make() seems to be not working:
```
wburns, line 17: What isn't working exactly?

manager/src/main/java/io/gingersnapproject/json/JsonHelper.java
fax4ever: Maybe we could avoid parsing the JSON hits, building the entire QueryResult JSON manually from a StringBuilder. But in any case let's keep it as it is for now.

manager/src/main/java/io/gingersnapproject/search/QueryHandler.java
fax4ever: Much better thanks. I'm applying it.

manager/src/main/java/io/gingersnapproject/json/JsonHelper.java
fax4ever: If I try:
``` java
// Json.make() seems to be not working:
// > doing it manually
Json result = Json.make(queryResult);
```

I get the error:
``` bash
Don't know how to convert to Json : QueryResult[hitCount=1, hitCountExact=true, hits=[{"name":"BA0666","scheduled_time":"12:00:00","airline_id":{"iata":"BA","name":"British Airways"},"gate_id":{"name":"B1"}}], hitsExacts=true]
```

fax4ever:
Thank @wburns I applied the changes.

ryanemerson:
Thanks @fax4ever

PR #91, opened 2023-02-27
Ensure the DatabaseHandler finishes initialization before the CacheRu…

ryanemerson:


manager/src/main/java/io/gingersnapproject/database/DatabaseHandler.java
```diff
@@ -1,53 +1,50 @@
...
-   @ConfigProperty(name = "quarkus.datasource.db-kind")
-   String dbKind;
    Vendor vendor;
 
    Map<String, Table> tables = Collections.emptyMap();
    Map<String, String> table2rule = Collections.emptyMap();
 
-   void start(@Observes  @Priority(10) StartupEvent ignore) {
-      vendor = Vendor.fromDbKind(dbKind);
+   DatabaseHandler() {}
```
rigazilla: do we also need the default const?

ryanemerson: It seems not, removed :slightly_smiling_face:

wburns:
This smells like something else to me, but I also don't want to hold up the demo, so this is fine with me.

